Issue #7, opened 2015-11-12
XDG compatibility


zabbal:
Right now settings and binaries are stored in ~/.xmonad/
It would be great if XDG specification is respected and ~/.config/xmonad/ is used instead. Other generated files (like xmonad.hi) could go to $XDG_CACHE_HOME/xmonad/
See http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html for the reference.

Delapouite:
This move has been done by neovim recently. Glad to see this initiative being more widespread.

zabbal:
Note: there's been some patches circulating in https://code.google.com/p/xmonad/issues/detail?id=484 but their current whereabouts is unknown.

pjones:
If someone wants to submit a PR I'd be happy to review it.

And changes should definitely prefer to use `~/.xmonad` if it exists, otherwise it could use the XDG layout.  That would allow for backwards compatibility.

zsedem:
I submitted a pull request on this 
#26

pjones:
This is related to #53 and #54.  Once those issues are resolved I think something like this could be moved into xmonad-contrib.

pjones:
Fixed in #62

PR #26, opened 2016-02-27
#7 XDG Compatibility

zsedem:
add support for configure from ~/.config/xmonad folder

### DIFF
--- a/CONFIG
+++ b/CONFIG
@@ -1,8 +1,10 @@
 == Configuring xmonad ==
 
-xmonad is configured by creating and editing the file:
+xmonad is configured by xmonad.hs the paths searched are:
 
     ~/.xmonad/xmonad.hs
+    $XDG_CONFIG_DIR/xmonad/xmonad.hs
+    ~/.config/xmonad/xmonad.hs
 
 xmonad then uses settings from this file as arguments to the window manager,
 on startup. For a complete example of possible settings, see the file:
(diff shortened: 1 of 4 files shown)


CONFIG
```diff
@@ -1,8 +1,10 @@
 == Configuring xmonad ==
 
-xmonad is configured by creating and editing the file:
+xmonad is configured by xmonad.hs the paths are searched are:
```
liskin: drop the first "are"

geekosaur:
I still want to see proper XDG support, not chuck-all-the-files-under-$XDG_CONFIG_HOME which violates XDG and is not compatible with network home directories.

Additionally, there's a number of open issues that all touch the same xmonad core code and really need to be done together, unless we plan to rewrite that code completely for each change. See https://github.com/geekosaur/xmonad.hs/blob/master/new-startup-config for an outline that addresses all of:
- XDG (and proper OS X --- yes, people use xmonad on OS X) support
- sandbox support (cabal and stack)
- a proper fix for the xmonad-on-$PATH issue, plus other mechanisms to work around XDG sessions essentially requiring a full session manager
- more flexibility in order to support e.g. testing configurations (this is not discussed in the above but the design is actually based around working code I've used for that; I still need to work out interactions with the sandbox support)
- handling the xmessage problem (this part is still subject to change, as the current spec is very definitely a hack based around accommodating existing contribs' usage of xmessage instead of rewriting them)

Some of these actually conflict rather badly, notably naïve sandboxes are incompatible with XDG. I don't expect to be able to resolve that, but I can make it possible for the user to resolve it if they need to instead of forcing them into noncompliance.

PR #29, opened 2016-03-18
help: correct section for workspace switching

siimphh:
Wouldn't expect this sectioning to be intentional, no?

Also, it looks like vim really wanted to add a newline to eof which appears consistent with the other nearby files so I kept it.


geekosaur:
Applied, thanks.

PR #50, opened 2016-10-11
Add support for recompiling via stack

mgsloan:
This adds support for recompiling xmonad via stack, allowing for control over the environment xmonad gets built in. Here's how this works:

If there is a `stack.yaml` file next to `xmonad.hs`, then it uses it for recompilation.  This is as simple as using `stack ghc`, pointing it at the stack.yaml file, and providing it the same arguments normally given to `ghc`.


colonelpanic8:
This is awesome! Any chance this gets merged any time soon?

mgsloan:
It seems more likely that it will instead allow use of a custom build script - https://github.com/xmonad/xmonad/pull/53

PR #52, opened 2016-10-16
Add finallyX.

Fuco1:
Run a job in the X monad and then run cleanup even in case of exception.

Useful for when I want to clean up some global state after executing an action.  An analogy to `finally` from IO exceptions.

### DIFF
--- a/src/XMonad/Core.hs
+++ b/src/XMonad/Core.hs
@@ -23,7 +23,7 @@ module XMonad.Core (
     Layout(..), readsLayout, Typeable, Message,
     SomeMessage(..), fromMessage, LayoutMessages(..),
     StateExtension(..), ExtensionClass(..),
-    runX, catchX, userCode, userCodeDef, io, catchIO, installSignalHandlers, uninstallSignalHandlers,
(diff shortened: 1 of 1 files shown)


pjones:
I wonder if this would go better in an `Errors` module in xmonad-contrib.  There are a lot of exception helpers and I'm not sure where we should draw the line when it comes to `X` wrappers for them.

Anyone else have any thoughts?

geekosaur:
On Thu, Dec 8, 2016 at 6:51 PM, Peter J. Jones <notifications@github.com>
wrote:

> I wonder if this would go better in an Errors module in xmonad-contrib.
> There are a lot of exception helpers and I'm not sure where we should draw
> the line when it comes to X wrappers for them.
>
Unless there's a clear use case in the core itself, I'd put it under
XMonad.Util.

-- 

Fuco1:
I figured since there is `runX` and `catchX` I would put it there to be discoverable, but I can open a MR elsewhere.  Is `XMonad.Util` in contrib?

I use this function in one patch I have lined up for contrib (so far not submitted as it depends on this being available).

pjones:
@Fuco1 Yes, `XMonad.Util` is in contrib.  Thanks.

PR #56, opened 2016-11-08
Fallback onto /etc/xmonad

NeQuissimus:
I would like to place my `xmonad.hs` in a global folder (outside `~/.xmonad`). This patch would allow for the file to exist in `/etc/xmonad` instead.

The patch does not have any impact (from what I can tell) on existing XMonad configurations.
It merely offers a fallback mechanism, should `~/.xmonad` not exist (other than `def`).

This relates to NixOS/nixpkgs#20260

### DIFF
--- a/src/XMonad/Core.hs
+++ b/src/XMonad/Core.hs
@@ -431,9 +431,12 @@ runOnWorkspaces job = do
              $ current ws : visible ws
     modify $ \s -> s { windowset = ws { current = c, visible = v, hidden = h } }
 
--- | Return the path to @~\/.xmonad@.
+-- | Return the path to @~\/.xmonad@ if it exists, otherwise use @\/etc\/xmonad@.
 getXMonadDir :: MonadIO m => m String
-getXMonadDir = io $ getAppUserDataDirectory "xmonad"
+getXMonadDir = io $ do e <- userExists; if e then user else (return "/etc/xmonad")
+  where
+  user = getAppUserDataDirectory "xmonad"
+  userExists = user >>= doesDirectoryExist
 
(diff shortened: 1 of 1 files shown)


pjones:
I'm also a NixOS user and will be working on putting my xmonad config/binary in the Nix store.  I don't think having xmonad use `/etc/xmonad` is a good idea, especially on a multiple-user system.

After I get my xmonad working out of `/nix/store` I'll post back here with a (hopefully) better solution.

geekosaur:
On Thu, Dec 8, 2016 at 6:06 PM, Peter J. Jones <notifications@github.com>
wrote:

> I'm also a NixOS user and will be working on putting my xmonad
> config/binary in the Nix store. I don't think having xmonad use
> /etc/xmonad is a good idea, especially on a multiple-user system.
>

I'd prefer a variant: ability to copy a bootstrap config from there (or,
better, stock in /usr/share/xmonad with local overrides in /etc/xmonad).
That said, there are some specific use cases (kiosk configurations, for
example) for a single global config with no user customization. But... I'm
not sure xmonad is the droid you are looking for in those cases.


-- 

pjones:
I'm going to close this PR since this feature will be available once #62 is merged.

@NeQuissimus for your usecase, just make sure the following environment variables are set before starting xmonad:

```sh
export XMONAD_CONFIG_DIR=/etc/xmonad
export XMONAD_CACHE_DIR=/etc/xmonad
export XMONAD_DATA_DIR=/etc/xmonad
``` 

(Or tailor those variables to your liking.  The latter two should probably be in /var.)

PR #62, opened 2016-12-13
Support for custom config locations and recompile strategies

pjones:
This PR collects up all of the commits that I'm making for issue #61.  It makes more sense to work in a PR since we have the ability to do code reviews here.


pjones:
I've finished my testing and believe this is ready to merge.

If you have a moment, please review the code.  Thanks.

/cc @geekosaur @dmwit @byorgey

CHANGES.md
```diff
@@ -1,8 +1,76 @@
...
+
+      2. XDG Base Directory Specification directories, if they exist:
+
+         - `XDG_CONFIG_HOME/xmonad`
+         - `XDG_CACHE_HOME/xmonad`
+         - `XDG_DATA_HOME/xmonad`
+
+      3. The `~/.xmonad` directory.
+
+    If none of these directories exist then (1) will be used if it is
```
byorgey: I don't understand what is meant by "If none of these directories exist" here.  Is it correct?

CHANGES.md
```diff
@@ -1,8 +1,76 @@
...
+    exists in the xmonad configuration directory it will be called
+    instead of `ghc`.  It takes one argument, the name of the
+    executable binary it must produce.
+
+    Note: the build script is called whenever `ghc` would have been
+    called.  That means the `xmonad.hs` file (or any files in `lib`
+    need to have newer time stamps than the generated executable in
+    order for the build script to be called.
+
+    This is one possible fix for issue #8.
```
byorgey: This kind of makes it sound like we haven't decided which of two fixes we are going to use for #8.  But if I understand correctly, what you really mean is that there are now two mechanisms by which a user can solve the problem described in #8.  Is that correct?  If so, maybe a better way to phrase it would be something like "This is one of two mechanisms that can be used to achieve sandboxed xmonad builds (see #8)"?

CHANGES.md
```diff
@@ -1,8 +1,76 @@
...
+    `xmonad`.
+
+    This will keep xmonad from using its configuration file
+    checking/compiling code and directly start the window manager
+    without `exec`ing any other binary.
+
+    See the documentation for the `launch` function in `XMonad.Main`
+    for more details.
+
+    This is another possible solution to issue #8.
```
byorgey: Same comment as above.

src/XMonad/Core.hs
```diff
@@ -438,39 +440,124 @@ runOnWorkspaces job = do
              $ current ws : visible ws
     modify $ \s -> s { windowset = ws { current = c, visible = v, hidden = h } }
 
--- | Return the path to @~\/.xmonad@.
+-- | Return the path to the xmonad configuration directory.  Several
```
byorgey: Since there are now multiple xmonad directories (configuration, cache, data), it would be good to include here a quick description of what should go in the config directory, and under what circumstances it will be accessed.  I am not sure I completely understand this myself.

CHANGES.md
byorgey: Oh, I think I understand now, you mean if none of them exist then (1) *will be created* if it is set, otherwise (2) will be created?  I think the text ought to mention something about directories being created.

src/XMonad/Core.hs
```diff
@@ -438,39 +440,124 @@ runOnWorkspaces job = do
...
-
--- | 'recompile force', recompile @~\/.xmonad\/xmonad.hs@ when any of the
--- following apply:
+getXMonadDir =
+    findFirstDirWithEnv "XMONAD_CONFIG_DIR"
+      [ getXdgDirectory XdgConfig "xmonad"
+      , getAppUserDataDirectory "xmonad"
+      ]
+
+-- | Return the path to the xmonad cache directory.  Several
```
byorgey: Same comment as above re: clarifying purpose of cache directory.

src/XMonad/Core.hs
```diff
@@ -438,39 +440,124 @@ runOnWorkspaces job = do
...
+-- directories exist then (1) will be used if it is set, otherwise (2)
+-- will be used.  Either way, a directory will be created if necessary.
+getXMonadCacheDir :: MonadIO m => m String
+getXMonadCacheDir =
+    findFirstDirWithEnv "XMONAD_CACHE_DIR"
+      [ getXdgDirectory XdgCache "xmonad"
+      , getAppUserDataDirectory "xmonad"
+      ]
+
+-- | Return the path to the xmonad data directory.  Several
```
byorgey: Same comment as above re: clarifying purpose of data directory.

src/XMonad/Main.hs
```diff
@@ -186,24 +193,12 @@ xmonadNoargs initxmc serializedWinset serializedExtstate = do
                  return (fromMaybe fbc_ v)
 
     hSetBuffering stdout NoBuffering
+    xinesc <- getCleanedScreenInfo dpy
```
byorgey: What was the reason for moving this line down?  I don't really know what it does so I'm just curious.

src/XMonad/Operations.hs
```diff
@@ -442,20 +443,78 @@ initColor dpy c = C.handle (\(C.SomeException _) -> return Nothing) $
...
+    path  <- stateFileName
+    stateData <- gets (\s -> StateFile (wsData s) (extState s))
+    catchIO (writeFile path $ show stateData)
+
+-- | Read the state of a previous xmonad instance from a file and
+-- return that state.
+readStateFile :: (LayoutClass l Window, Read (l Window)) => XConfig l -> X (Maybe XState)
+readStateFile xmc = do
+    path <- stateFileName
+    raw  <- userCode $ io (readFile path <* removeFile path)
```
byorgey: Does this work properly with lazy IO?  I'm afraid it will try to remove the file before actually reading it.

byorgey: In any case I don't understand why it needs to be removed at all.  Seems better to just let it hang around until it is overwritten on the next restart.  That way if xmonad crashes the state can still be restored from the most recent restart.

byorgey (approved):
Generally looks good to me.  I just have a few questions/suggestions.

src/XMonad/Main.hs
pjones: I didn't mean to move this line down.  It was an artifact of sloppy refactoring.  I'll move it back.

pjones:
@byorgey I updated my branch based on your suggestions.  This should be ready to merge now.  If I don't hear from you in the next couple of days I'll just merge this.  Thanks.

byorgey:
Looks good to me, thanks!

PR #64, opened 2016-12-14
fix #63: window jumping to origin position when dragging

### LINKED ISSUE #63: Dragging a window that moves other windows in response to being moved is bugged
When you drag the main window of an application that has custom code to move other windows belonging to the same application (for example tool windows) in response to the move event on the main window, then the main window will jump between the drag target point and its original location.

A real-world application where this occurs is the Android Emulator:

emulator

If you move the main window, the toolbar window will automatically be moved as well so that it is always next to the main window. 

If you try to do this in XMonad, what happens is that you cannot move the emulator's main window, because when you try to drag it, it will start to jump between the where the mouse is (drag target position) and its original position (where it was before the dragging started).

bennofs:
We need to make sure that the StackSet contains the right
position for the window at all times while dragging. Previously,
the window was only placed at a different position. If, for any
reason, a layout refresh happens while the window is being dragged,
then the window will jump back to the old position because the StackSet
still contains the position of the window where it was as the drag started.

This patch fixes that issue by calling float after each mouse movement, which
ensures that the position for the window is updated in the StackSet.

Fixes #63

### DIFF
--- a/src/XMonad/Operations.hs
+++ b/src/XMonad/Operations.hs
@@ -515,8 +515,11 @@ mouseMoveWindow w = whenX (isClient w) $ withDisplay $ \d -> do
     (_, _, _, ox', oy', _, _, _) <- io $ queryPointer d w
     let ox = fromIntegral ox'
         oy = fromIntegral oy'
-    mouseDrag (\ex ey -> io $ moveWindow d w (fromIntegral (fromIntegral (wa_x wa) + (ex - ox)))
-                                             (fromIntegral (fromIntegral (wa_y wa) + (ey - oy))))
+    mouseDrag (\ex ey -> do
+                  io $ moveWindow d w (fromIntegral (fromIntegral (wa_x wa) + (ex - ox)))
+                                      (fromIntegral (fromIntegral (wa_y wa) + (ey - oy)))
+                  float w
+              )
               (float w)
 
 -- | resize the window under the cursor with the mouse while it is dragged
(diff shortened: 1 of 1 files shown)


bennofs:
I have a small (well, it uses Qt5) test program that replicates the issue at https://github.com/bennofs/repros/tree/master/xmonad/%2363 (build using `qmake && make`)

pjones:
@bennofs Do we need this same change made to window resizing with the mouse?

bennofs:
@pjones That's a good point, let me extend my test application to do something on resize.

bennofs:
@pjones you were right, we also need to do this when resizing. fixed.

byorgey:
Looks good to me, thanks!

yusi1:
Issue seems to be occurring again for me (same thing happens on every application)

liskin:
@yusi1 When you say again, what changed? Did you upgrade xmonad or perhaps just change your config file?

yusi1:
Yes I did update XMonad, this hasn't been happening for the past two weeks while using my config

liskin:
> Yes I did update XMonad, this hasn't been happening for the past two weeks while using my config

Would you happen to remember from which version did you upgrade and what version you have now?

yusi1:
Currently I am on XMonad 0.15-122, unfortunately I do not remember the version I was on before

yusi1:
Turns out this issue wasn't any window managers fault, it was my compositor, sorry for the misunderstanding.

PR #65, opened 2016-12-25
Don't send focus events for the wrong windows

HebaruSan:
I'm seeing an occasional focus switching loop after rapid mouse movements when UpdatePointer is used with focusFollowsMouse=True. After moving the pointer back and forth between two windows, or a linear movement across the corners of three, the pointer flickers very quickly between two of the windows until I press super-tab. It looks like UpdatePointer is reacting to an event that's one frame behind the actual mouse pointer.

I found this patch on XMonad's old Google Code site (see below), which adds an additional sanity check to make sure the pointer is still in the window before generating the focus event, and after testing it the problem seems to be fixed.

Originally from https://code.google.com/archive/p/xmonad/issues/200 comment 2


pjones:
@HebaruSan According to the other comments on that issue, this patch didn't fix the issue.  You are saying that it "seems" to be fixed.  Can you reliably reproduce this issue on master without this patch applied?

PR #76, opened 2017-02-09
Rare crash fix due to uncaught exception from `getWindowAttributes`

SOwOphie:
### Description

I was getting some very strange crashes which I could only reproduce reliably on my desktop machine (but e.g. not on my laptop which is running a very similar setup), so I'm still not entirely sure what exactly triggered it. However, after much headscratching, I managed to track the error message down to this `getWindowAttributes` call which doesn't check for exceptions. The attached patch adds a minimal "sane default" for when the call fails, which fixed the crashing for me.

### Checklist

  - [ ] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing) _Given the nature of the issue and of my changes, I don't think using various configs from `xmonad-testing` will provide any meaningful results, but I'll happily catch up on that if requested._

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -69,6 +69,8 @@
     Fixes #8.  (Second way to have a custom build environment for
     XMonad.  See previous entry for another solution.)
 
+  * Fixed a crash related to an unhandled getWindowAttributes exception
+
 ## 0.12 (December 14, 2015)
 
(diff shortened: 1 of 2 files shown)


pjones:
Thanks!

SOwOphie:
Wow, that was fast! My pleasure! Woo, first PR successful =)

PR #91, opened 2017-03-30
Remove the xmonad state file after reading it

### LINKED ISSUE #86: Serialized state is persistent after reboot
Because the state of xmonad is now serialized to a file, my xmonad breaks sometimes because it found the serialized state from the previous session and i end up having to remove the file and restart xmonad.

Maybe it's a good idea to remove the statefile after it has been read?
Or put it somewhere like /run/user/$(id -u) or /tmp?

pjones:
### Description

Remove the xmonad state file after reading it

Tries to make sure IO is not lazy so the file is processed before it is removed from the file system.

Fixes #86 and friends.

### Checklist

  - [X] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

### DIFF
--- a/src/XMonad/Operations.hs
+++ b/src/XMonad/Operations.hs
@@ -465,14 +465,21 @@ writeStateToFile = do
     catchIO (writeFile path $ show stateData)
 
 -- | Read the state of a previous xmonad instance from a file and
--- return that state.
+-- return that state.  The state file is removed after reading it.
 readStateFile :: (LayoutClass l Window, Read (l Window)) => XConfig l -> X (Maybe XState)
 readStateFile xmc = do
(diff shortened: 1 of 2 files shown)


src/XMonad/Operations.hs
```diff
@@ -491,6 +498,9 @@ readStateFile xmc = do
                            [(x, "")] -> Just x
                            _         -> Nothing
 
+    readStrict :: Handle -> IO String
```
peti, line 501: It would probably be much easier to just use [Data.ByteString.readFile](http://hackage.haskell.org/package/bytestring-0.10.8.2/docs/Data-ByteString.html#v:readFile), which is strict.

PR #110, opened 2017-05-05
Safer string quoting for help message

yyny:
Using `show` to quote help string instead of hard-coding it.
Allows for quotes and other characters to be placed inside the help string.

### Description

Include a description for your changes, including the motivation
behind them.

> Ran into this after copying this code to add my own help message using [data-embed](https://hackage.haskell.org/package/data-embed) (You guys should consider embedding the help message aswell. Not sure if compile time file embedding is worth the price though (Extra dependency + TemplateHaskell).
> It took me a good few minutes to find the problem.

### Checklist

  - [ x ] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [ x ] I've confirmed these changes don't belong in xmonad-contrib instead

  - [x] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

> Not worth it, where clause might be indented wrong and not sure how documentation comments work in this situation but everything should be OK
> I just made this change from within the GitHub website.

  - [ ] I updated the `CHANGES.md` file

> Not worth it

PR #111, opened 2017-05-08
Safer string quoting for help message

yyny:
Using `show` to quote help string instead of hard-coding it.
Allows for quotes and other characters to be placed inside the help string.

See #110.

### DIFF
--- a/src/XMonad/Config.hs
+++ b/src/XMonad/Config.hs
@@ -221,9 +221,9 @@ keys conf@(XConfig {XMonad.modMask = modMask}) = M.fromList $
     , ((modMask .|. shiftMask, xK_q     ), io (exitWith ExitSuccess)) -- %! Quit xmonad
     , ((modMask              , xK_q     ), spawn "if type xmonad; then xmonad --recompile && xmonad --restart; else xmessage xmonad not in \\$PATH: \"$PATH\"; fi") -- %! Restart xmonad
 
-    , ((modMask .|. shiftMask, xK_slash ), spawn ("echo \"" ++ help ++ "\" | xmessage -file -")) -- %! Run xmessage with a summary of the default keybindings (useful for beginners)
+    , ((modMask .|. shiftMask, xK_slash ), helpCommand)
     -- repeat the binding for non-American layout keyboards
-    , ((modMask              , xK_question), spawn ("echo \"" ++ help ++ "\" | xmessage -file -"))
+    , ((modMask              , xK_question), helpCommand)
     ]
     ++
     -- mod-[1..9] %! Switch to workspace N
@@ -237,6 +237,9 @@ keys conf@(XConfig {XMonad.modMask = modMask}) = M.fromList $
     [((m .|. modMask, key), screenWorkspace sc >>= flip whenJust (windows . f))
         | (key, sc) <- zip [xK_w, xK_e, xK_r] [0..]
         , (f, m) <- [(W.view, 0), (W.shift, shiftMask)]]
+  where
+    helpCommand :: X ()
+    helpCommand = spawn ("echo " ++ show help ++ " | xmessage -file -")  -- %! Run xmessage with a summary of the default keybindings (useful for beginners)
 
 -- | Mouse bindings: default actions bound to mouse events
 mouseBindings :: XConfig Layout -> M.Map (KeyMask, Button) (Window -> X ())


erthalion (commented):
I'm wondering how these changes from line 187 to 222 are related to the subject of this PR? If it's just different code formatting maybe it makes sense to avoid them or (if it's important) suggest within a different PR, what do you think?

yyny:
The style change is actually a seperate commit, I'm not sure you noticed this by now, you didn't seem to realise it in your review comment.
Anyway, here is what I have to say. It will get a bit lengthy, so I have split it into multiple sections. Some might be of more importance to you and the project than others.

**Reasoning behind second commit**
It's not up to me to decide if the commit can stay, however, I made the style change because while making my changes for the first commit, it became really apparent to me that line 226 had alignment that differed from the rest, something which wasn't quite as obvious when the line was long and complicated (Which is probably why other contributers missed it). Now seemed like a good time to change this, especially since this file in particulair hasn't been updated in a year acording to it's history.

**Opinion about a second pull request**
Once again, the decision is not up to me, however, IMHO multiple pull requests would be overdo. I would personally prefer a rebase of each individual commit the way it is done for projects with a more organized and well defined pull request protocol (for lack of a better word) like nodejs, however, this does not seem to be the protocol in this repository, probably for good reasons (I can think of some).

There are three options here:

  * Merge as a single pull request
    * The pushed commits will not include information about how and why the style was updated. This very conversation justifies most of that though.
  * Rebase as a single pull request
    * The pushed commits will include all information we want, without an additional unnecesairy merge commit, in fact. However, the pushed changes will not line up with the rest of the merged pull requests so far, and might even be considered unfair to them, as these new rebased commits would contain more information than the old merged commits would. The rebased commit would also be attributed differently. (Also because it will create multiple small commits instead of one large one).
  * Merge as multiple pull requests
    * Would create a pull requests for what is actually just a style update related to the change in my first commit. IMHO not worth it. In that case I would rather run the entire source directory through a linter instead. My linter came up with several changes I refrained from committing.

**Opinion about importance of style change and similair kinds of changes**
Like I mentioned earlier, I personally don't consider the style change on it's own to be important enough to deserve the overhead of a seperate pull request. However, I consider every inconsistency in a codebase to be a problem, including inconsistent formatting. If the formatting doesn't get updated now this inconsistency might stay there for several years to come.
As I have noticed by now, the problem with these kinds of inconsistency fixes on GitHub or any multi-contributor environment in general is that the sheer amount of communicating required to advertise and pull the change far outweighs the benefits that will ever be had from it.
However, I continue to insist that every inconsistency in a project should be fixed, however small it might be.
The irony here is that this entire pull request is built on this philosophy; If I didn't create this pull request, chances are noone ever would, however, I spend several hours figuring out why my XMonad environment suddenly stopped working as indented to what in the end came down to a small inconsistency noone else would have ever noticed, and that in itself is enough of a reason for me to create the puil request anyway.

**Amending pull request**
Line 205 still didn't get updated correctly. I will see if there is a way I can amend this without destroying the pull request (Like I have done before, I would love if someone could explain to me how these kinds of changes interfere with GitHub Pull Requests).

pjones:
As a general rule we don't accept PRs that only contain style changes due to the way they change file history without introducing actual functional changes.  I don't see a compelling argument why changing code alignment is worth committing.

yyny:
@pjones Thanks for the clarifications. Should I remove the second commit then, as my PR does _not_ just contain style changes? (If that's even possible, which I don't think it is. At this point, I'm fine with someone else commiting the change, as long as it doesn't require yet another new pull request)

ggreif:
No need for new PR,, just (force-)push to the existing branch.

yyny:
@ggreif Thanks! That worked.

pjones:
Thanks!

schlueter:
The force push seems a bit extreme as now the entire conversation here is missing an important piece of context. Would it be possible for someone who has an original copy of the branch to post the patch text of the (now omitted) style change?

On a related note, is the rationale as to not accepting style change PRs elaborated upon anywhere?

In both cases, I'm mostly curious, but answers would make it clearer to future potential contributors.

pjones:
@schlueter:

If you rebase or otherwise modify a commit that you have made public, using `--force-with-lease` is the correct thing to do.  I don't think it's an extreme thing to do, and there is enough context in these comments to understand what happened.

The reason we want the rebase/force is to keep the repository history clean.  It doesn't make any sense to merge in a set of commits where later commits revert earlier commits.  We would much rather merge in the actually changes that need to be made and completely avoid other superfluous commits.  As you mentioned, a patch-based system doesn't have this issue.  The other potential solution is to perform a squashed merged.

PRs that only change code formatting would need to be reviewed on a case-by-case basis.  Generally, changes that only modify the formatting of code make it harder to understand the history of a file, especially when you want to study how a single function changed over time.  Formatting/styling can be so subjective that I think it would be hard to justify this type of PR.

PR #127, opened 2017-08-06
remove the man pages from data-files

dudebout:
The man pages are available for packagers in `extra-source-files`.

Having them in `data-files` is confusing since, according to Cabal's user guide [1], `data-files` contains "A list of files to be installed for run-time use by the package.", but the man pages are not used at run-time by xmonad.

[1]: https://www.haskell.org/cabal/users-guide/developing-packages.html

### Description

behind them.

### Checklist

### DIFF
--- a/xmonad.cabal
+++ b/xmonad.cabal
@@ -34,8 +34,6 @@ tested-with:
                     GHC==7.10.3,
                     GHC==8.0.1
 
-data-files:         man/xmonad.hs, man/xmonad.1, man/xmonad.1.html
-
 source-repository head
   type:     git
   location: https://github.com/xmonad/xmonad


jpotier:
Also waiting on review for this patch. Maybe @geekosaur or @pjones can get a look at it? I'm sorry for pushing, but it's a blocking problem downstream, and the patch is trivial (or so it seems)

jpotier:
Hmmm, building patched version of xmonad seems to fail:
```
$ nix-repl '<nixpkgs>'
:b xmonad.overrideAttrs (drv: { patches = [ (builtins.fetchurl "https://github.com/dudebout/xmonad/commit/3dc65c3d2ecbe5aff5883aba80b682f1128bd43e.patch";) ]})
```
with
```
mv: missing destination file operand after '/nix/store/gaamhsy5bic01llfnmgri1vf5ykdlfvb-xmonad-0.13-doc/share/man/man1/'
Try 'mv --help' for more information.
```

dudebout:
@jpotier this is expected. nixpkgs is currently picking up the man page from `data-files`. We need a fix in cabal2nix to pick up the man page from the source directory and not from the runtime dependency directory: see https://github.com/NixOS/cabal2nix/pull/290.

jpotier:
So if this patch is accepted, it'll break downstream at rebuild, right?

peti:
Installing man pages as "data files" is wrong and IMHO you should merge this patch regardless of what it will do to downstream simply because it fixes incorrect / misleading behavior. Nixpkgs and other will be more than happy to remove the workarounds have that were previously added to cope with the current installation scheme when the new version comes out.

PR #137, opened 2017-09-14
Use precise in travisCI for now

colonelpanic8:
This fixes the build for now. We should figure out why things don't work in trusty though.


geekosaur:
When they moved to precise, quite a few projects had to request additional packages be added to their image. Presumably this starts again with trusty.

peti:
I re-generated the `travis.yml` script with https://github.com/haskell-CI/haskell-ci and it's working fine now.

PR #153, opened 2018-02-06
update to work with latest GHC

vmchale:
### Description

This just makes every monoid instance into a semigroup instance. 

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -18,6 +18,8 @@
     the focus follows when the mouse enters another workspace
     but not moving into any window.
 
+  * Compiles with GHC 8.4.0
+
 ## 0.13 (February 10, 2017)
 
 ### Breaking Changes
--- a/xmonad.cabal
+++ b/xmonad.cabal
@@ -1,5 +1,5 @@
 name:               xmonad
-version:            0.14
+version:            0.13
 homepage:           http://xmonad.org
 synopsis:           A tiling window manager
 description:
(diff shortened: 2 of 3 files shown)


MichielDerhaeg:
If I recall correctly, the lowest version of GHC that we support is 7.6.3. The `Data.Semigroup` module was added with GHC 8.0. So I'm afraid you can't fix it this way.

MichielDerhaeg:
Simply adding
```
if !impl(ghc >= 8.0)
  build-depends: semigroups == 0.18.*
```
might be the best solution. If everyone agrees to add another dependency.

MichielDerhaeg:
Nevermind i tested it and it works with older versions like this.

MichielDerhaeg:
@byorgey could you merge?

xmonad.cabal
```diff
@@ -32,7 +32,8 @@ tested-with:
                     GHC==7.6.3,
                     GHC==7.8.4,
                     GHC==7.10.3,
-                    GHC==8.0.1
+                    GHC==8.0.1,
+                    GHC==8.2.2
```
byorgey, line 36: Shouldn't we add `GHC==8.4.1` as well?

CHANGES.md
```diff
@@ -18,6 +18,8 @@
     the focus follows when the mouse enters another workspace
     but not moving into any window.
 
+  * Compiles with GHC 8.4.0
```
byorgey, line 21: This should say `8.4.1` now that it has actually been released.

PR #157, opened 2018-03-21
Travis

byorgey:
Regenerate `.travis.yml` to include latest innovations in multi-ghc-travis technology, and also to make sure we are actually testing `xmonad` on GHC 8.2 and 8.4 with travis.

The only thing I'm slightly unsure about is what goes wrong if we remove that `profiling` flag in `xmonad.cabal`.  `cabal check` complained about it and we don't seem to be using it anywhere, but I could be wrong.


peti:
The `-prof` flag should definitely be removed from the Cabal file. The proper way to enable profiling is to set `--enable-profiling` during the `cabal configure` stage. What we currently have with `-fprofiling` is basically a re-implementation of a mechanism that `cabal` already has.

I'll make the necessary changes *after* releasing 0.14, just to avoid confusing people who might rely on this behavior.

Issue #123, opened 2017-07-07
generatemanpage requires pandoc 1.19 to build


clinty:
util/GenerateManpage.hs treats writerTemplate as a Maybe String.  This API change occurred in pandoc 1.19, and using older pandoc results in things like

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=867386

Please express a lower bound on pandoc 1.19 for the generatemanpage stanza

clinty:
It also appears to be incompatible with pandoc 2.1

PR #165, opened 2018-04-25
Port GenerateManpage.hs to pandoc 2

clinty:
Closes: #123

There is a regression here in terms of aesthetics.


MichielDerhaeg:
Don't forget to add ``text`` as a dependency in ``xmonad.cabal`` as well when you import it's modules.

clinty:
Confusing that it built for me without that.

peti:
Merged.

PR #167, opened 2018-05-10
Log information about xmonad compile + avoid unnecessary recompile

mgsloan:
### Description

Particularly with the addition of build scripts, it can be tricky to figure out
what XMonad is doing when attempting recompilation.  This makes it clearer by
adding some logging.

Due to this logging, I noticed that the lag of xmonad start was because it was
always recompiling!  When I startup my computer, I do not want it to delay
rebuilding my window manager. This also fixes that issue such that it only
recompiles XMonad if it is going to reinvoke due to getProgName not being the
expected string.

### Checklist

      - [x] I have tested these changes with [my own xmonad configuration](https://github.com/mgsloan/dotfiles). Btw, is it possible that the above checklist item might reduce contribution?  How about making it optional?  I'd hope that xmonad-testing would be involved in some sort of CI, so contributors should not need to do that manually.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -23,6 +23,18 @@
   * Restored compatability with GHC version prior to 8.0.1 by removing the
     dependency on directory version 1.2.3.
 
+  * xmonad no longer always recompile on startup. Now it only does so if the
+    executable does not have the name that would be used for the compilation
+    output. The purpose of recompiling and executing the results in this case is
+    so that the `xmonad` executable in the package can be used with custom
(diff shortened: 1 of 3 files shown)


mgsloan:
CI failures seem unrelated to the changes

mgsloan:
FYI I have another change atop this on my master branch, that allows for overriding the recompilation function https://github.com/mgsloan/xmonad/commit/cead3a1e3ece8312d188d08e51ef421465822a27

PR #168, opened 2018-05-12
Fix some warnings

mgsloan:
Can't fix all import warnings, since travis tests for ghc < 7.10

### Description

behind them.

### Checklist

### DIFF
--- a/src/XMonad/Core.hs
+++ b/src/XMonad/Core.hs
@@ -551,13 +551,13 @@ getXDGDirectory :: XDGDirectory -> FilePath -> IO FilePath
 getXDGDirectory xdgDir suffix =
   normalise . (</> suffix) <$>
   case xdgDir of
-    XDGData   -> get "XDG_DATA_HOME"   ".local/share"
-    XDGConfig -> get "XDG_CONFIG_HOME" ".config"
-    XDGCache  -> get "XDG_CACHE_HOME"  ".cache"
+    XDGData   -> getEnvPath "XDG_DATA_HOME"   ".local/share"
+    XDGConfig -> getEnvPath "XDG_CONFIG_HOME" ".config"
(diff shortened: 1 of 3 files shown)


src/XMonad/Main.hs
```diff
@@ -1,3 +1,4 @@
+{-# OPTIONS_GHC -fno-warn-warnings-deprecations #-} -- For migrateState
```
peti, line 1: I think it's actually counter-productive to silence those warnings. The proper solution is to change code that relies on deprecated features or functions to use the newer and supported solutions instead.

src/XMonad/Operations.hs
```diff
@@ -1,4 +1,4 @@
-{-# OPTIONS_GHC -fno-warn-orphans #-}
+{-# OPTIONS_GHC -fno-warn-orphans -fno-warn-warnings-deprecations #-}
```
peti, line 1: ditto

peti (changes requested):
Adding `-fno-warn-warnings-deprecations` feels like the wrong direction to go. Can you please remove those two changes from the PR?

src/XMonad/Core.hs
```diff
@@ -551,13 +551,13 @@ getXDGDirectory :: XDGDirectory -> FilePath -> IO FilePath
 getXDGDirectory xdgDir suffix =
   normalise . (</> suffix) <$>
   case xdgDir of
-    XDGData   -> get "XDG_DATA_HOME"   ".local/share"
```
peti, line 554: Which warning does this solve, exactly? Can you please cut & paste an example in the commit message?

peti:
I think this PR needs a bit of an update to be mergeable, and it's not clear whether that update is forthcoming. I'll close it for the time being just for the sake of keeping our list of pending PRs nice and tidy.

PR #173, opened 2018-07-05
Fix manpage generation

MichielDerhaeg:
Fixed the default config so manpage generation works.
Credit to @clinty for porting the utility to pandoc 2 (includes #165)


peti:
Merged.

PR #176, opened 2018-07-31
Reimplement sendMessage to deal properly with windowset changes made during handling

LSLeary:
**NB**: this PR probably shouldn't go into any minor version of 0.14 being made any time soon, and really should be reviewed by someone knowledgeable in xmonad's internals. I believe that the change is harmless, but I might be wrong.

### Summary

`XMonad.Operations`
 * Add `windowBracket`: provide a means of handling windowset changes
   made during the course of arbitrary `X` actions. Buys composability.
 * Add `windowBracket_` variant.
 * Add `modifyWindowSet` utility for use with the above.
 * Re-implement `sendMessage` using `windowBracket_` so that its refresh
   handles changes made to the windowset by the message handler.

### Motivation

Certain layouts like `XMonad.Layout.Groups` (and also one I've been using / writing-off-and-on for months) need to make changes to the stack (e.g. which window has focus) in message handling. Currently, `sendMessage` runs `windows` *after* passing in the message, however windowset changes won't be handled properly if `windows` can't see both the initial and the final state. Consequently the message handlers do naughty and messy things like running `windows` themselves or manually dealing with window border colours (and hoping no other changes need special handling). xmonad/xmonad-contrib#191 is a bug that seems to be resulting from the former strategy.

It is feasible to write in contrib this `sendMessage` as an alternative version `sendMessage'` and have layouts with this problem require its use in order for messages to be handled properly. However this violates a basic expectation, places a burden on the end user, and further encourages the problematic practice of baking message handling into exported actions (at the cost of composability).

By replacing the standard `sendMessage` instead, we can automatically roll out the extended handling to almost all messages being sent. Further, since the purpose of the refresh in `sendMessage` is to institute changes made in message handling, the fact that not all changes are being properly handled could really be considered a bug.

### Checklist

    - See arguments above.  

    - Tested as my system xmonad instance.

### DIFF
--- a/src/XMonad/Operations.hs
+++ b/src/XMonad/Operations.hs
@@ -21,7 +21,7 @@ import XMonad.Layout (Full(..))
 import qualified XMonad.StackSet as W
 
 import Data.Maybe
-import Data.Monoid          (Endo(..))
+import Data.Monoid          (Endo(..),Any(..))
 import Data.List            (nub, (\\), find)
 import Data.Bits            ((.|.), (.&.), complement, testBit)
 import Data.Ratio
@@ -30,6 +30,7 @@ import qualified Data.Set as S
 
 import Control.Applicative((<$>), (<*>))
 import Control.Arrow (second)
+import Control.Monad (void)
 import Control.Monad.Reader
 import Control.Monad.State
 import qualified Control.Exception.Extensible as C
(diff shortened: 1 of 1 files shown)


byorgey:
Just FYI, "someone knowledgeable in xmonad's internals" is not me (at least not anymore).  Maybe @geekosaur can take a look?

LSLeary:
@byorgey OK, I've been using this branch on my system for a long time now without any issues—lets just merge it. I imagine it'll get more than enough testing from other people before there's another major version anyway.

byorgey:
OK!

PR #177, opened 2018-08-02
[WIP] prepare for GHC 8.6.1

vmchale:
### Description

Updated XMonad to use the latest version of the `unix` package, allowing it to work with GHC 8.6.1. This requires the `unix` package to be uploaded to hackage so it will not pass CI until then.

### Checklist

### DIFF
--- a/xmonad.cabal
+++ b/xmonad.cabal
@@ -71,7 +71,7 @@ library
                    setlocale,
                    mtl,
                    process,
-                   unix,
+                   unix >= 2.8.0.0,
                    utf8-string >= 0.3 && < 1.1,
                    X11>=1.8 && < 1.10,
                    semigroups
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## unknown (unknown)
 
+### Bug Fixes
+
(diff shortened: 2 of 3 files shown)


xmonad.cabal
```diff
@@ -71,7 +71,7 @@ library
                    setlocale,
                    mtl,
                    process,
-                   unix,
+                   unix >= 2.8.0.0,
```
peti, line 74: I don't think that depending on the latest version is a viable solution. `unix` is a core library of `ghc` and as such cannot be updated easily.

hvr: Note that unix-2.8.0.0 doesn't exist yet, nor has its API been finalised (some early GHC 8.6.1 alphas shipped with a premature snapshot of unix-2.8) GHC 8.6.1 is going to ship with unix-2.7 

PS: The upcoming reinstallable `lib:ghc` planned for GHC 8.8 will finally remove the technical limitation that prevented from updating packages such as `unix`.

peti:
425c3c087260b00777f71631eea38edd9dad6a92 fixes the build with GHC-8.6.1 and adds it to the list of supported compilers that ought to be verified on travis-ci.

Issue #5, opened 2015-09-26
Floating window resize issue


mathstuf:
So making a summary for this isn't easy. Here's the setup:
- two monitors
- different resolutions (important!)
- keyboard focus on A
- cursor on B

When using a manage hook to float dialogs (`isDialog --> doFloat`), it is stored in the layout according to the cursor screen resolution (B). When the window actually appears, it shows up next to the currently focused window (on A). However, because the resolutions are different, the ratio stored in XMonad for its location is a different size. When the window appears, XMonad resizes it to be the same _ratio_, but that means the _resolution_ is different. Dialogs (possibly made worse due to improper size hints) then either get inflated with empty space or squished together with overlapping widgets.

The fix should be to determine the ratio of new windows based on where they will appear rather than the cursor-current screen resolution.

See coldfix/udiskie#79.

geekosaur:
I had already noticed this and was considering fixes, but as there haven't been complaints previously I was only considering windows with a defined aspect ratio or resize increment (e.g. most terminals, the programs inside which can break rather badly).

mathstuf:
It also appears that XMonad isn't respecting size hints about minimum or maximum sizes, but that's probably a different issue.

schlueter:
This was originally reported to the xmonad mailing list [back in 2010](http://haskell.1045720.n5.nabble.com/Issue-414-in-xmonad-Wrong-floating-window-sizes-on-multimon-setup-td5803294.html), and although they had some insights not mentioned here, that thread didn't go anywhere either.

One notable insight:

>I think the problem is that xmonad internally represents window dimensions  
as fractions of screen size, and these are computed from the initial window  
position and size. So, if an application creates a new window at (0,0) and  
then asks the window manager to position it, xmonad computes the size as a  
fraction relative to the screen containing (0,0) and then places the window  
on whatever screen is currently active. If the size of this screen is  
different, so is the size of the placed window.
-- Massive Rhino

bluss:
This seems to be the same bug: kupferlauncher/kupfer/issues/74

jonhoo:
@pjones is this something that is being worked on? I keep seeing this pop up all over the place, though mostly for gtk dialog boxes (file pickers, launcher windows, pinentry-gtk), and it is quite distracting. Is it something you can reproduce locally?

pjones:
@jonhoo I don't believe this is actively being worked on, there just aren't enough people contributing code.  If you have some time I would love to see a patch ;)

[...]

PR #197, opened 2019-07-25
Use current workspace to set dimensions of unmapped floating windows

Narthorn:
This fixes a bug when using multiple screens with different dimensions, causing floating windows to always increase in size or always decrease in size every time they're opened.

In such a situation, a floating window is first mapped at (0,0) (on the leftmost screen), using the dimensions of that screen, before being inserted in the current workspace, which may be on another screen, causing the window to be resized if that screen has larger or smaller dimensions. If the window then saves its size in pixels when closed, the next time it opens, it will go through this process again, causing it to become even larger or smaller.

This commit fixes this by using the current workspace to calculate dimensions for unmapped floating windows, instead of the workspace at screen position (0,0).

---

However, the same kind of problem remains if, for example, a custom ManageHook is used to automatically move a floating window to a workspace on a screen with different dimensions than the current workspace. I'm not sure what the proper fix should be for that, short of rewriting the entire floating layer to use pixel dimensions instead of "percent-of-screen-size" dimensions.

---

### Checklist

### DIFF
--- a/src/XMonad/Operations.hs
+++ b/src/XMonad/Operations.hs
@@ -565,7 +565,10 @@ floatLocation w =
           ws <- gets windowset
           wa <- io $ getWindowAttributes d w
           let bw = (fromIntegral . wa_border_width) wa
-          sc <- fromMaybe (W.current ws) <$> pointScreen (fi $ wa_x wa) (fi $ wa_y wa)
+          managed <- isClient w
+          sc <- if managed
+                then fromMaybe (W.current ws) <$> pointScreen (fi $ wa_x wa) (fi $ wa_y wa)
+                else return (W.current ws)
 
           let sr = screenRect . W.screenDetail $ sc
               rr = W.RationalRect ((fi (wa_x wa) - fi (rect_x sr)) % fi (rect_width sr))


liskin:
This seems to fix https://github.com/xmonad/xmonad/issues/5 indeed. I cherry-picked, restarted and will be running with the patch, reporting any issues should I find any.

src/XMonad/Operations.hs
```diff
@@ -565,7 +565,10 @@ floatLocation w =
...
           wa <- io $ getWindowAttributes d w
           let bw = (fromIntegral . wa_border_width) wa
-          sc <- fromMaybe (W.current ws) <$> pointScreen (fi $ wa_x wa) (fi $ wa_y wa)
+          managed <- isClient w
+          sc <- if managed
+                then fromMaybe (W.current ws) <$> pointScreen (fi $ wa_x wa) (fi $ wa_y wa)
+                else return (W.current ws)
 
           let sr = screenRect . W.screenDetail $ sc
               rr = W.RationalRect ((fi (wa_x wa) - fi (rect_x sr)) % fi (rect_width sr))
```
liskin, line 574: The above now computes negative position coordinates on all screens other than the one with `(wa_x wa, wa_y wa)` (usually 0, 0). This is then fixed by `adjust` in `manage`, but other uses of `floatLocation` (`doFloat`, …) do not make those adjustments. This means that a `doFloat` in manage hook will result in the window being placed at 0, 0 even though it belongs to a workspace on another screen.

To fix this, we must not use the screen rect if we didn't use pointScreen to obtain the correct screen rect in the first place. Perhaps just place the float at the center. I'll try making this change and will continue testing with that.

Also, we might consider moving the `adjust` function here.

liskin: https://github.com/liskin/xmonad/commit/207173d83ad107f8e610ceff300d0a1618fae13e

Narthorn:
I don't think this fixes #5 totally. It just assumes any floating window wants to end up on the currently active workspace, which should be the most common case, but might not always be.
So the problem is basically shifted to a less common case, while the most common case is handled correctly.

geekosaur:
The other case can't be fixed, because you need to run the layout to know where any window on a workspace will be placed, and that's only safe for visible workspaces or if the layout happens to support `pureLayout` properly.

I have wished several times that `pureLayout` were required to produce a reasonable approximation of how a workspace would be laid out (this would make `EwmhDesktops` work better with pager panel applets, for one), but we don't currently. (I *think* you currently get `Full` if it's not implemented, as opposed to getting an exception or etc.)

src/XMonad/Operations.hs
Narthorn: Good catch. I was using `doCenterFloat` instead of `doFloat`, so that issue didn't occur for me. 
I don't think `adjust` should go straight up in `floatLocation` however, as there are still times where you want windows to be able to go beyond screen boundaries (i.e. when dragging a window from screen to screen with the mouse, float is called repeatedly, making the window teleport around unpleasantly when it's crossing screens.)

liskin: Oh, right, dragging does indeed behave erratically now. I'll try something else later today.

liskin: How about this: https://github.com/liskin/xmonad/commit/9bcb0c3219108bc33ccdfb20edf9a181a0208dc5 ?

I moved `adjust` back to `manage` and changed `floatLocation` to only adjust when the window is unmanaged and its x,y coordinates are outside of current screen. This should fix the flickering and should also stop forcing float centering of windows that happen to be placed in the correct screen. (And I added explicit centering to my managehook as I want that behaviour anyway.)

liskin: Another fix for flickering of dragging windows, this time when dragging them outside screen: https://github.com/liskin/xmonad/commit/48b0caa613922124ffcadb2b112fcd1d6018061d

liskin:
@Narthorn I've been running https://github.com/xmonad/xmonad/compare/master...liskin:48b0caa613922124ffcadb2b112fcd1d6018061d for the last two months without encountering any additional problems. Would you like to pull my fixups into your PR, or shall I perhaps create my own PR and credit your for the initial idea?

Narthorn:
Sorry, I've meant to come back to this a while ago. If I recall correctly there's a bit of weirdness left, such as different behavior between leftmost screen and other screens when running doFloat (windows on leftmost screen spawn at (0,0), while on other screens they spawn in the center), and windows being teleported to center when unmanaged/remanaged while leftmost corner is outside screen boundaries, although that last case is a bit niche.

liskin:
@Narthorn Yeah, the first weirdness is definitely there, and it's easy to work around by using a `placeHook` or something, which turns out to be a good idea anyway (gpg pinentry ends up at 0,0 even without the patch; ssh-askpass uses framebuffer center regardless of xinerama layout; …). But it shouldn't be difficult to change the calculation to try to use the location somehow. It's not going to matter in my setup, but making things less weird in default configuration is probably worth it.

The second one I'm not sure how to reproduce.

liskin:
I've decided it's time to clean this up and submit it for merging, so here it is: https://github.com/xmonad/xmonad/pull/221

liskin:
#221 was just merged so we can close this one.

PR #200, opened 2019-10-30
Make ScreenDetail a newtype and RationalRect strict in its contents

ghost:
This seems to be the original intention. Should not have much of an effect but
could help with edge-case space leaks.

### Description

The change to ScreenDetail should not have much of an effect, but if there is one it should be an improvement.

The change to RationalRect will force all changes in dimensions to be calculated when they are 'made', rather than when they are used. As long as the calculations are simple, this should be a net improvement even if not all of the values are used.

Overall this is more of a 'cleanup' than an improvement.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -29,6 +29,9 @@
     In the cases of 1. and 3., the build script or executable is
     expected to be in the config dir.
 
+  * Change `ScreenDetail` to a newtype and make `RationalRect` strict in
+    its contents.
+
 ## 0.15 (September 30, 2018)
 
   * Reimplement `sendMessage` to deal properly with windowset changes made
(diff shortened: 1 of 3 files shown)


slotThe:
This looks nice to me, @wygulmage do you mind rebasing on top of current `master`?

ghost:
> This looks nice to me, @wygulmage do you mind rebasing on top of current `master`?

Should be good now.

slotThe:
@psibi I think this looks good, what do you say?

psibi (approved):
Looks okay to me. I will merge this in a couple of days unless anybody has any objection.

ghost:
Just for the record, I've been using this change for over a year with no problems.

psibi:
Thank you!

PR #202, opened 2019-11-10
Add Functor, Foldable, and Traversable instances for StackSet.Stack

ghost:
### Description
Add some basic class instances to the zipper type.

'transformers' is added as a dependency in xmonad.cabal (to get Data.Functor.Reverse), but it was already a transitive dependency of mtl.

### Checklist

### DIFF
--- a/src/XMonad/StackSet.hs
+++ b/src/XMonad/StackSet.hs
@@ -1,4 +1,5 @@
 {-# LANGUAGE PatternGuards #-}
+{-# LANGUAGE DeriveFunctor #-}
 
 -----------------------------------------------------------------------------
 -- |
@@ -52,6 +53,8 @@ module XMonad.StackSet (
     ) where
 
 import Prelude hiding (filter)
+import Control.Applicative.Backwards (Backwards (Backwards, forwards))
+import Data.Foldable (foldr, toList)
 import Data.Maybe   (listToMaybe,isJust,fromMaybe)
 import qualified Data.List as L (deleteBy,find,splitAt,filter,nub)
 import Data.List ( (\\) )
@@ -175,8 +178,19 @@ data RationalRect = RationalRect !Rational !Rational !Rational !Rational
 data Stack a = Stack { focus  :: !a        -- focused thing in this set
                      , up     :: [a]       -- clowns to the left
                      , down   :: [a] }     -- jokers to the right
-    deriving (Show, Read, Eq)
+    deriving (Show, Read, Eq, Functor)
 
+instance Foldable Stack where
+    toList = integrate
+    foldr f z = foldr f z . toList
+
+instance Traversable Stack where
+    traverse f s =
+        flip Stack
+            -- 'Backwards' applies the Applicative in reverse order.
+            <$> forwards (traverse (Backwards . f) (up s))
+            <*> f (focus s)
+            <*> traverse f (down s)
 
 -- | this function indicates to catch that an error is expected
 abort :: String -> a
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -35,6 +35,8 @@
   * Fixed dunst notifications being obscured when moving floats.
     https://github.com/xmonad/xmonad/issues/208
 
+  * Added `Foldable`, `Functor`, and `Traversable` instances for `Stack`.
+
 ## 0.15 (September 30, 2018)
 
   * Reimplement `sendMessage` to deal properly with windowset changes made
--- a/stack.yaml
+++ b/stack.yaml
@@ -6,6 +6,13 @@ packages:
 extra-deps:
 - X11-1.9.2
 
+flags:
+  xmonad:
+    # stack doesn't support automatic flags
+    # https://cabal.readthedocs.io/en/latest/cabal-package.html#resolution-of-conditions-and-flags
+    # https://github.com/commercialhaskell/stack/issues/1313#issuecomment-157259270
+    quickcheck-classes: false
+
 nix:
   packages:
     - zlib
--- a/tests/Properties.hs
+++ b/tests/Properties.hs
@@ -196,6 +196,5 @@ tests =
   ,("pointWithin",        property prop_point_within)
   ,("pointWithin mirror", property prop_point_within_mirror)
 
-  ]
-
-
+  ] <>
+  prop_laws_Stack
--- a/tests/Properties/Stack.hs
+++ b/tests/Properties/Stack.hs
@@ -1,13 +1,19 @@
 {-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE GeneralizedNewtypeDeriving #-}
 module Properties.Stack where
 
 import Test.QuickCheck
+import Test.QuickCheck.Classes (
+    Laws (lawsTypeclass, lawsProperties), Proxy1 (Proxy1),
+    foldableLaws, traversableLaws,
+    )
 import Instances
 
 import XMonad.StackSet hiding (filter)
 import qualified XMonad.StackSet as S (filter)
 
 import Data.Maybe
+import Data.Functor.Classes (Eq1 (liftEq), Show1 (liftShowsPrec))
 
 
 -- The list returned by index should be the same length as the actual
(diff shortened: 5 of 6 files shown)


ghost:
Is there a standard way to deal with these pandoc build failures in CI?

src/XMonad/StackSet.hs
```diff
@@ -178,6 +180,23 @@ data Stack a = Stack { focus  :: !a        -- focused thing in this set
     deriving (Show, Read, Eq)
 
 
+instance Functor Stack where
+    fmap f s = Stack
+        (f (focus s))
+        (fmap f (up s))
+        (fmap f (down s))
```
slotThe: This looks `derive`able

src/XMonad/StackSet.hs
```diff
@@ -178,6 +180,23 @@ data Stack a = Stack { focus  :: !a        -- focused thing in this set
...
+instance Foldable Stack where
+    toList = integrate
+    foldr f z = foldr f z . toList
+
+instance Traversable Stack where
+    traverse f s =
+        flip Stack
+            <$> fmap getReverse (traverse f (Reverse (up s)))
+            <*> f (focus s)
+            <*> traverse f (down s)
```
slotThe: I can see that these are law abiding, but it would perhaps be nice to also tell this to `QuickCheck` :)

slotThe (changes requested):
This looks good, just thought it worth mentioning a few things I've noticed

EDIT: There's also the obligatory "Please don't merge `master` into your branch, but rebase instead", but all of that can be fixed quite easily at the end :)

src/XMonad/StackSet.hs
ghost: I don't know QuickCheck well enough to test ``forall (t :: Applicative n => m a -> n a)  f. t . traverse f = traverse (t . f)`` or ``forall g f. traverse (Compose . fmap g . f) = Compose . fmap (traverse g) . traverse f``.

I could write a test for ``traverse Identity = Identity``, which would be telling.

src/XMonad/StackSet.hs
ghost: I didn't want to add DeriveFunctor to the module, but if that's OK then I 100% agree.

slotThe: I think if something can be derived, it should be; I'm sure the xmonad source code can take one more language extension ;)

src/XMonad/StackSet.hs
slotThe: You're right, this is probably not as straightforward as I thought it would be.  

The easiest way, I suppose, would be to depend on [some library](https://hackage.haskell.org/package/quickcheck-classes-0.6.4.0/docs/Test-QuickCheck-Classes.html#v:traversableLaws) that has already solved this problem; I don't think this would be much of an issue, as it'd be solely in the test-suite.

Then again, we're really only sticking together known 'Foldable' and 'Traversable's, so I'm not sure if this is worth it after all.  @liskin @byorgey @psibi opinions?

liskin: I wouldn't insist on having the instance tested, personally, but that's proably because I don't remember being bitten by a non-law-abiding instance. The coding style section of XMonad.Doc.Developing states that all pure functions must have quickcheck properties, so if one can be added relatively easily, it'd be nice. `quickcheck-classes` isn't in lts-12, however, so it may be somewhat painful (probably less painful than pandoc, though, so probably tolerable? I have no idea any more).

ghost:
It's now using DeriveFunctor, and I added a test for `traverse Identity`.

Could use `foldMapDefault` to check that `toList` via `traverse` gives the same result as `integrate` (which it does), but I haven't added that test yet.

Edit: Added the updated xmonad.cabal file, which now exposes transformers.
Edit 2: Removed unused 'foldl' import to make the GHC 8.4.4 check succeed.

src/XMonad/StackSet.hs
slotThe: >I wouldn't insist on having the instance tested, personally, but that's proably because I don't remember being bitten by a non-law-abiding instance. 

I have to confess that I was at one point; maybe that's why I'm so scared about this :) 

I do, however, think that if we _do_ decide to test for the properties in question, we should test for _all_ of them, and not only a few selected ones (probably with the help of some library)

slotThe:
@wygulmage you should rebase on top of `master` instead of merging it into your feature branch; otherwise we will get a very hard-to-follow history

ghost:
> @wygulmage you should rebase on top of `master` instead of merging it into your feature branch; otherwise we will get a very hard-to-follow history

Sorry, that was my bad. I'll fix it.

src/XMonad/StackSet.hs
ghost: So should I just remove the identity test? I understand the importance of testing, but for something like this, honestly, testing is _not good enough_. It should be _provably_ correct.

slotThe: I mean, sure I can prove that the instances are correct, but short of using something like liquid haskell (can it do this even?) I can't really convince GHC of this...  I think you should remove the identity test in every case though, yes.

Two ways forward that I can see

  1. Test _everything_ with the use of some library (additional dependencies _shouldn't_ matter because they would only be compiled on a `stack test` or equivalent)
  2. Test nothing because the instances are "obvious" (they are, but so are a few of the pure functions and yet they are still quickchecked, as `XMonad.Doc.Extending` demands)

@geekosaur do you have an opinion on this?

geekosaur: In the past, verification has been done with proof assistants rather than Haskell-based testing. This, for example, is why the `StackSet` takes so many type variables when they're only ever instantiated to fixed types: they were instantiated differently in (iirc) Coq since the X11 types weren't available. That doesn't mean we're forced to go the same way now, but it is a data point.

(I'm not sure we even have the scaffolding for the old tests, which could probably stand to be re-run. Someone would have to ask @donsbot what happened to them, since I don't think they're in the repo.)

In the meantime, simple testing is probably worth doing but it's not worth trying to get too fancy; if we can't reasonably test something in Haskell then it's not worth bothering.

slotThe: Alright, @wygulmage I think it's then worth going for option 1. and use a library to fully quickcheck these instances.

ghost: OK. I have updated the tests to use quickcheck-classes (and removed the 'ad-hoc' `Identity` test). This was much easier than I expected. I had to add an `Arbitrary` instance for `Stack`, which I am slightly less confident in than the `Traversable` instance.

To make the tests compile with GHC 8.4.4, I had to add `Eq1` and `Show1` instances for `Stack`. I am _much_ less confident in the `Show1` instance, but I don't know if it can be properly derived. I'm not sure if it matters, because nothing else depends on it. I could use a newtype for testing and define the instances on that to avoid exporting them for `Stack`.

slotThe: > To make the tests compile with GHC 8.4.4, I had to add `Eq1` and `Show1` instances for `Stack`. I am _much_ less confident in the `Show1` instance, but I don't know if it can be properly derived. I'm not sure if it matters, because nothing else depends on it. I could use a newtype for testing and define the instances on that to avoid exporting them for `Stack`.

I think that we should either move these into `test` (they would become orphans at that point, but since it's a testing repository this may or may not matter) or use the afforementioned newtype

geekosaur: Orphans are acceptable in executables, including tests.

ghost: I've switched the tests and `Arbitrary`, `Show1`, and `Eq1` instances to a `TestStack` newtype in Properties/Stack.hs.

slotThe (approved):
Wonderful!

From my side the only thing that's left to be done is to squash a few commits that don't really make sense now (like the one about changing out `Reverse` with `Backwards` and the ones about making `Stack` an instance of a certain class which is now moved to `tests`)

ghost:
Is it OK to squash it down to 1 commit, or do you want some of the relevant history preserved?

slotThe:
On Wed, Feb 03 2021 12:49, wygulmage wrote:
> Is it OK to squash it down to 1 commit, or do you want some of the
> relevant history preserved?

That's up to you really; personally, I'd also be fine with one commit
for adding the instances and one for the tests or similar things

slotThe:
@liskin I think this is mergeable now; what do you think?

ghost:
I haven't updated CHANGES.md. If you want that done before merging I can do it now. I was holding off because this has gone through a lot of revisions.

liskin:
I think it's probably ready to merge but I'm not comfortable merging this before the CI is fixed and we can test against various LTS versions again. I'm afraid there's little we can do here apart from waiting for the situation to resolve. Please make yourselves comfortable, this might take a few weeks/months. I'm sorry.

ghost:
No worries! Honestly after the first year I didn't expect this PR to go anywhere.

liskin:
(I pushed a rebase to resolve a conflict and trigger the CI, which is now fully operational again.)

liskin:
So apparently there's no quickcheck-classes in LTS-12. What to do:

* stop supporting (testing against) LTS-12
* add cabal flag to make the quickcheck-classes dependency optional and ifdef the affected tests
* any better ideas?

slotThe:
On Wed, Mar 31 2021 07:15, Tomáš Janoušek wrote:
> * any better ideas?

What about just adding it to `extra-deps'?  Will that cause any breakage
that I'm not seeing right now?

liskin:
> What about just adding it to `extra-deps'? Will that cause any breakage that I'm not seeing right now?

It might break some constraints when we add newer LTS versions later but at that point we'll probably be okay with dropping LTS-12 (and thus dropping the extra-deps), so this could be a good way forward, provided it builds. :-)

liskin:
> provided it builds

It doesn't :-(
https://github.com/wygulmage/xmonad/runs/2241141631?check_suite_focus=true

slotThe:
On Wed, Mar 31 2021 15:17, Tomáš Janoušek wrote:
>> provided it builds
>
> It doesn't :-(
> https://github.com/wygulmage/xmonad/runs/2241141631?check_suite_focus=true

Sigh.  Okay then I'm also out of ideas.  I feel like both CPP and not
testing on LTS-12 are ugly options, but perhaps CPP is the lesser of the
two evils?

ghost:
> On Wed, Mar 31 2021 15:17, Tomáš Janoušek wrote: > provided it builds It doesn't :-( https://github.com/wygulmage/xmonad/runs/2241141631?check_suite_focus=true
> Sigh. Okay then I'm also out of ideas. I feel like both CPP and not testing on LTS-12 are ugly options, but perhaps CPP is the lesser of the two evils?

In principle, ugly is better than untested. But since the instances _have now been tested_, this is really just about ensuring that changes to the instances will also be tested. The instances don't depend on anything that should change from computer to computer.

liskin:
Agreed that it's not really a big deal. I'd say ease of maintenance is more important here than obsessing about having the generated instances tested in all possible configurations.

liskin:
I pushed some commits to fix the builds, but I can't resist feeling that this whole thing wasn't worth the time, as what I ended up with is even uglier than I expected. :-/

liskin:
Anyway, I'd prefer not wasting more time on this, so unless someone else comes up with something else that results in "All checks have passed", I'll merge this.

src/XMonad/StackSet.hs
```diff
@@ -175,8 +178,19 @@ data RationalRect = RationalRect !Rational !Rational !Rational !Rational
...
+    deriving (Show, Read, Eq, Functor)
+
+instance Foldable Stack where
+    toList = integrate
+    foldr f z = foldr f z . toList
+
+instance Traversable Stack where
+    traverse f s =
+        flip Stack
+            -- 'Reverse' applies the Applicative in reverse order.
```
slotThe: This should be `'Backwards'`

slotThe (approved):
Seems fine to me, but then again I get a headache every time I start to look into build systems... I just found a small doc remnant from when this still used `Data.Functor.Reverse`

src/XMonad/StackSet.hs
liskin: Good catch.

PR #203, opened 2019-11-16
Update example config link

owm111:
The link for the example config leads to file saying "go here instead"; change initial link to "there".

### Description

Well, save a few clicks whenever you go looking for an example config because you forgot how configs work (speaking from experience).

### Checklist

    > It's a tiny change to the README, could it break stuff?

    > Not yet.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,5 +125,5 @@ For a program dispatch menu:
 [xmonadcontrib]: https://hackage.haskell.org/package/xmonad-contrib
 [xmc-prompt-shell]: https://hackage.haskell.org/package/xmonad-contrib/docs/XMonad-Prompt-Shell.html
 [platform]: http://haskell.org/platform/
-[example-config]: https://github.com/xmonad/xmonad-testing/blob/master/example-config.hs
+[example-config]: https://github.com/xmonad/xmonad-contrib/blob/master/XMonad/Config/Example.hs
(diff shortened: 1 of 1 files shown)


psibi:
Thanks!

geekosaur:
The link to the old "config file" should go away entirely / be replaced by the new one. The old was never a good idea and led to a bunch of bad configs that resist xmonad updates; the double link was basically backward compatibility which has outlived its usefulness. The original intent is better handled by good documentation instead of by a bad config file.

psibi:
@geekosaur What do you think is the new one ? There seems to be quite some config there.

geekosaur:
The huge config is an example of overriding every configuration option… and does so by overriding with some particular version's defaults, which can cause breakage in new (e.g. when dmenu's usage changed) or conversely old (the help keybinding did this since the function it was bound to didn't exist) versions. It should really be done with documentation since it's not that understandable, and it's asking for trouble when people use it as a "default config". The page with a pair of very short configs (one using XMonad.Config.Desktop) and examples of additions should be the real landing page, and the 0.9- config page should at this point probably go away or at least not be linked to directly any more.

Note that we've had Linux distributions file off the "DO NOT USE THIS CONFIG!" comment at the top and recommend or even install for users the bad config as a default, and this has (as with the dmenu change) caused breakage in the past.

Issue #81, opened 2017-02-15
Add option to create borders externally


incertia:
### Problem Description

Borders are created internally, so if borderWidth = 2, a call to xcb_create_window() with width=1280 and height=720 and borderwidth = 0 ends up creating a client area measuring 1276 x 716. For graphics applications that require a fixed window size, this can be particularly annoying if you have to wait for the window to become exposed to resize the window.

An .xinitrc file consisting of precisely

```
#! /bin/bash
exec xterm
```

can create borders that go outward [like so](http://i.imgur.com/Ycs51pV.png), confirmed via xcb_get_geometry having the correct client area.

### Configuration File

```haskell
module Main (main) where

import XMonad

main :: IO ()
main = xmonad def
```

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [x] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

pjones:
@incertia You seem to know way more about this than I do.  Any chance you want to submit a patch?

erthalion:
@pjones @incertia if you don't mind I'll try to tackle this question.

pjones:
@erthalion Got for it!

erthalion:
> An .xinitrc file consisting of precisely ... can create borders that go outward like so

Well, I'm really confused by this statement. I couldn't find any mention of
external/outside borders in `libx11` source code, `xcb_get_geometry_reply_t`
(a structure related to the `xcb_get_geometry`) also doesn't contain any
information, that may help to detect whether borders are external or internal.

I couldn't also reproduce a situation when borders are outside of a window
using provided `.xinitrc` and the xmonad config. So it would be nice if
@incertia will explain this point in more details.

> For graphics applications that require a fixed window size

So, at this point I see following options:

* I'm wrong and it's possible to create borders outside of a window, so we can
  use this feature.

* "Emulate" external borders by creating a window with dimensions `window_width +   2 * border_width x window_heigth + 2 * border_width`. But it means some overhead
    for the operations with window sizes, because it's necessary to calculate
    it since we're pretending that dimensions are still `window_width x
    window_heigth`.

* It really makes sense only for graphics applications with fixes window size
  (I assume it's mostly about games), so it would be great to be able to
  differentiate those apps and just remove borders for them (but I almost sure
  that it's impossible).

[...]

Issue #198, opened 2019-08-31
borderWidth incorporated with window size hints (x11 or xmonad?)


mneyrane:
### Problem Description

When running applications, some floating windows (e.g. dialog or save prompts) are set to have a fixed size. Setting `borderWidth` to some non-zero value in `xmonad.hs` means that such windows (including the border) respect the size hint. This results in a "squished" window.

![squished window](https://0x0.st/z4vn.png)

I used LibreOffice to produce this prompt.

### Configuration File

Here is a minimal configuration to reproduce the issue.

```haskell
import XMonad

main = xmonad def
    { terminal    = "urxvt"
    , modMask     = mod4Mask
    , borderWidth = 2
    }
```

### Additional information

`xprop` output of save prompt:

```
WM_STATE(WM_STATE):
                window state: Normal
                icon window: 0x0
_NET_WM_STATE(ATOM) = _NET_WM_STATE_MODAL, _NET_WM_STATE_SKIP_TASKBAR
WM_HINTS(WM_HINTS):
                Client accepts input or input focus: True
                Initial state is Normal State.
                window id # of group leader: 0x3200001
_GTK_THEME_VARIANT(UTF8_STRING) = 
XdndAware(ATOM) = BITMAP
_NET_WM_OPAQUE_REGION(CARDINAL) = 0, 0, 456, 98
WM_TRANSIENT_FOR(WINDOW): window id # 0x3200024
_NET_WM_WINDOW_TYPE(ATOM) = _NET_WM_WINDOW_TYPE_DIALOG
_NET_WM_SYNC_REQUEST_COUNTER(CARDINAL) = 52432147, 52432148
_NET_WM_USER_TIME(CARDINAL) = 4389460
_NET_WM_USER_TIME_WINDOW(WINDOW): window id # 0x3200d12
WM_CLIENT_LEADER(WINDOW): window id # 0x3200001
_NET_WM_PID(CARDINAL) = 4150
WM_LOCALE_NAME(STRING) = "en_US.UTF-8"
WM_CLIENT_MACHINE(STRING) = "localhost"
WM_NORMAL_HINTS(WM_SIZE_HINTS):
                program specified location: 0, 0
                program specified minimum size: 460 by 102
                program specified maximum size: 460 by 102
                program specified base size: 0 by 0
                window gravity: NorthWest
[...]

PR #207, opened 2019-12-27
differentiate between floating and tiled in rects and only apply `tileWindow` to tiled windows

incertia:
This addresses #81. There's not much to test but if this totally kills the behavior of floating windows let me know and I'll see what other workarounds I can put in.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,9 @@
 
 ## unknown (unknown)
 
+  * Only apply `tileWindow` to tiled windows, so that floating windows are not
+    accidentally resized.
+
 ## 0.15 (September 30, 2018)
 
   * Reimplement `sendMessage` to deal properly with windowset changes made
--- a/src/XMonad/Operations.hs
+++ b/src/XMonad/Operations.hs
@@ -21,7 +21,7 @@ import XMonad.Layout (Full(..))
 import qualified XMonad.StackSet as W
 
 import Data.Maybe
-import Data.Monoid          (Endo(..),Any(..))
+import Data.Monoid          (Endo(..), Any(..), mconcat)
 import Data.List            (nub, (\\), find)
 import Data.Bits            ((.|.), (.&.), complement, testBit)
 import Data.Ratio
@@ -128,7 +128,7 @@ windows f = do
     -- for each workspace, layout the currently visible workspaces
     let allscreens     = W.screens ws
         summed_visible = scanl (++) [] $ map (W.integrate' . W.stack . W.workspace) allscreens
-    rects <- fmap concat $ forM (zip allscreens summed_visible) $ \ (w, vis) -> do
+    rects <- fmap mconcat $ forM (zip allscreens summed_visible) $ \ (w, vis) -> do
         let wsp   = W.workspace w
             this  = W.view n ws
             n     = W.tag wsp
@@ -151,11 +151,11 @@ windows f = do
 
         io $ restackWindows d (map fst vs)
         -- return the visible windows for this workspace:
-        return vs
+        return (flt, rs)
 
-    let visible = map fst rects
+    let visible = let (floating, tiled) = rects in map fst floating ++ map fst tiled
 
-    mapM_ (uncurry tileWindow) rects
+    mapM_ (uncurry tileWindow) (snd rects)
 
     whenJust (W.peek ws) $ \w -> do
       fbs <- asks (focusedBorderColor . config)
--- a/src/XMonad/Operations.hs
+++ b/src/XMonad/Operations.hs
@@ -155,6 +155,7 @@ windows f = do
 
     let visible = let (floating, tiled) = rects in map fst floating ++ map fst tiled
 
+    mapM_ (uncurry tileWindowFloating) (fst rects)
     mapM_ (uncurry tileWindow) (snd rects)
 
     whenJust (W.peek ws) $ \w -> do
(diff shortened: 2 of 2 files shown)


liskin:
Doesn't kill them completely, moving and resizing using the mouse still works, but other changes to window size/location seem to be ignored:

* [placeHook in manageHook](https://github.com/liskin/dotfiles/blob/4a7c860baeeb965d72157a0c6f49e1c152452eba/.xmonad/xmonad.hs#L217) is ignored
* [W.float in a key binding](https://github.com/liskin/dotfiles/blob/4a7c860baeeb965d72157a0c6f49e1c152452eba/.xmonad/xmonad.hs#L109) is ignored ad well

Both of these are expected, I guess, as you removed the call that is meant to sync xmonad's view of window placement with the X server.

incertia:
I can think about what to do with `W.float`. Maybe this means creating a version of `tileWindow` that does not affect the border and applying it to `flt`. `placeHook` should work as expected. The manage hook is composed somewhere in the `f` parameter of `windows` and is applied to the old set of windows and all further operations commence on the new value, `ws`, within the first few lines of `windows`.

wedens:
I've applied this patch, but some dialogs are still displayed like this:
image

When I set border width to 0, it displays dialog properly:
image

wedens:
Wait. For whatever reason recompiling and restarting xmonad was not enough. After reboot the problem is fixed (at least for this dialog).

liskin:
@wedens Reboot shouldn't be needed, but just a recompile/restart may not be enough on a non-traditional distro like NixOS. I'd look that way. (But this is off-topic here.)

geekosaur:
Mod-shift-space after restart?

On 6/11/20, Tomáš Janoušek <notifications@github.com> wrote:
> @wedens Reboot shouldn't be needed, but just a recompile/restart may not be
> enough on a non-traditional distro like NixOS. I'd look that way. (But this
> is off-topic here.)
>
> --
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/xmonad/xmonad/pull/207#issuecomment-642632754


-- 
brandon s allbery kf8nh
allbery.b@gmail.com

liskin:
> Mod-shift-space after restart?

On NixOS the binary may end up in an entirely different location as the path is a hash of the sources that produced it...

wedens:
Seems like this MR breaks moving floating windows with mod+left (default binding) mouse button. It moves the window and also resizes it.

liskin:
> Seems like this MR breaks moving floating windows with mod+left (default binding) mouse button. It moves the window and also resizes it.

That's a good point. Before this PR, `RationalRect` meant the same thing for all windows. After this PR, it suddenly means "outer window dimensions (incl border)" for tiled windows and "inner window dimensions (excl border)" for floating windows. So if you grab a tiled window and start moving it, it'll resize because its `RationalRect` suddenly means something else. And yeah I suppose it breaks moving floats as well because that uses `X.Operations.float` which invokes `floatLocation` and that adjusts for the border width, but now that `RationalRect` means something else that adjustment isn't good.

Anyway, I think a better fix for the issue (https://github.com/xmonad/xmonad/issues/198, rather than https://github.com/xmonad/xmonad/issues/81, because I don't think xmonad is ever going to become a reparenting WM that draws borders outside) would be to adjust the `RationalRect` in `X.Operations.manage`. That invokes `floatLocation` as well, but the border width correction doesn't actually happen because that window's border width is 0 at that point. We could tell `floatLocation` to correct for the future border width instead.

But then, we don't really know the border width is 0… it could well be something else if the window was created with a border. So perhaps an even better fix would be to adjust the floating `RationalRect` whenever the border width is changed. When [the border width is set initially during managing the window](https://github.com/xmonad/xmonad/blob/cde1a25bca9b7aeeb24af73588221f5f54ae770b/src/XMonad/Operations.hs#L304) we could get the previous border width, compare it to the border width being set, and adjust the floating `RationalRect`, if any. And the same thing at all the places in xmonad-contrib that change the border size.

It'll likely mess up fullscreen windows, especially games, somewhat, but I can't foresee how exactly. We'll have to try.

Another idea, possibly the simplest to implement, is to just add a manageHook that corrects for the difference in border widths. One could conditionally use this for ordinary floats and skip games/fullscreens as those won't end up having borders anyway.

liskin:
Hm, but all these ideas of mine seem impossible to make play nice with https://xmonad.github.io/xmonad-docs/xmonad-contrib/XMonad-Layout-NoBorders.html. That's a bit unfortunate. And yeah, with `RationalRect` meaning "inner window dimensions (excl border)" for floats this problem goes away (as long as every border width change is followed by a refresh that does moveResizeWindow)…

geekosaur:
Merge the Chrome hack to core? 😉

liskin:
> Merge the Chrome hack to core? 😉

That's a different issue. What I was talking about here is that X.L.NoBorders changes the border width a lot and the mechanics of whether a window ends up having a border or not are quite complex, so the only place we can reliably adjust the floating rect is X.L.NoBorders itself. But! All this border width adjustment happens in `runLayout`, and you can't mess with the floating rects in there, because `runLayout` is invoked from within `windows`.

We'd need to use something like https://xmonad.github.io/xmonad-docs/xmonad-contrib-0.18.0.9/XMonad-Util-ActionQueue.html (which was introduced for the implementation of https://xmonad.github.io/xmonad-docs/xmonad-contrib-0.18.0.9/XMonad-Hooks-BorderPerWindow.html, another border-related thingy, lol) to adjust the floating rects afterwards, which would double the already quite annoying flickering of X.L.NoBorders. And it's entirely possible this would result in endless loops, so we'd need a circuit-breaker somewhere.

It's just gets super ugly. I'll need to sleep on this.

liskin:
(suppose now's the time to look at @LSLeary's pull requests as some of the problems I'm anticipating here are probably solved by the design changes in those PRs… this got out of hand quite quickly 🙂)

PR #209, opened 2020-01-17
changed copyFromParent to 0 for the depth parameter of createWindow.

ghost:
In `XMonad.Main.replace`, `createWindow` was called with `copyFromParent` as
both the depth and the class.
Should only use it for the class, `copyFromParent` is not a depth.

I found this while changing X11 to use newtypes (for this exact reason).

### Description

behind them.

### Checklist

### DIFF
--- a/src/XMonad/Main.hs
+++ b/src/XMonad/Main.hs
@@ -506,7 +506,7 @@ replace dpy dflt rootw = do
             let screen = defaultScreenOfDisplay dpy
                 visual = defaultVisualOfScreen screen
                 attrmask = cWOverrideRedirect .|. cWEventMask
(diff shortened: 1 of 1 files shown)


liskin:
`XCreateWindow` manpage says:

> ```
> depth     Specifies the window's depth.  A depth of CopyFromParent means the depth is
>           taken from the parent.
> ```

so I'm not so sure it's wrong. But my knowledge of Xlib is quite limited.

ghost:
Thanks, liskin. I assumed it was a type mismatch because `depth` is signed and `class` is unsigned; I didn't realize it was another generic zero-value like `None`.

PR #217, opened 2020-03-22
Add Nix build support

epeery:
### Description

Project can now be built with ``nix-build release.nix`` and installed with ``nix-env --install ./result``

X dependencies are packaged so there is no need to manually install them.

Also removes the need to install/update Cabal leading to a potentially easier install experience.

> Note: [Xmobar does something similar
](https://github.com/jaor/xmobar/blob/master/default.nix)
### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Change Log / Release Notes
 
+## 0.15.1 (March 22, 2020)
+
+  * Added Nix build support
+
 ## unknown (unknown)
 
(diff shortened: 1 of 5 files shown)


peti (commented):
We don't want generated files checked into our repository. Also, pinning the build to a random version of Nixpkgs does more harm than good because we won't notice when the build fails after version updates of, say, the system dependencies.

I'd rather not merge this.

epeery:
Understandable. Thanks for taking the time to review this.

PR #219, opened 2020-03-31
Allow pattern-matching on 'XMonad.Layout.Choose' by exporting constructors

jumper149:
Export the constructor `Choose` to be able to construct and pattern-match `Choose l r a`. To do that it is also necessary to export `LR` with constructors.

### Description
I just added the constructor `Choose` and `LR` with constructors to the exports of the module `XMonad.Layout`

### Checklist

  - [ ] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing) (that seems to be a dead repository)


jumper149:
Here in [my dotfiles](https://github.com/jumper149/dotfiles/commit/9f43542e3cc1a5d622465cd2c3c2a5bada7a993f#diff-d420470135f162eccfc2922956e05a00R92-R115) you can see an example of what can be done with this change:

I use everything that `Choose` provides but wrap it in a newtype and define a different instance of `LayoutClass`,  so that I'm able to handle specific Messages differently. 
I send the messages `ModifyWindowBorderEnabled a` and `ModifyScreenBorderEnabled a` to both sub-layouts instead of just the current one.

This way I can set the spacing for all sub-layouts on my workspaces with just using [toggleGaps](https://github.com/jumper149/dotfiles/commit/9f43542e3cc1a5d622465cd2c3c2a5bada7a993f#diff-d420470135f162eccfc2922956e05a00R81-R87).

liskin:
I have no opinion about whether exposing these constructors should be done or not, but I think there is an easier way to implement this.

If I understand your setup correctly, you have a global toggle for gaps that broadcasts a message to all workspaces and all sublayouts and this message toggles the inner state of all those layouts (or layout modifiers). So what you really want is a single global boolean toggle instead of a hundred local ones that you toggle, hoping they'll never come out of sync. How about this:

Store this global toggle in [ExtensibleState](https://hackage.haskell.org/package/xmonad-contrib-0.16/docs/XMonad-Util-ExtensibleState.html) and whenever the layout is switched, sync this state into it using something like
```haskell
setGaps enabled :: X ()
setGaps enabled = do traverse_ broadcastMessage messages
￼                    refresh
￼   where messages = [ ModifyWindowBorderEnabled (const enabled)
￼                    , ModifyScreenBorderEnabled (const enabled)
￼                    ]
```

It's not the cleanest solution, I guess, but it's simpler than adding a LayoutClass that delivers messages to both active and inactive layouts (which is probably dangerous in general).

liskin:
I am now in favor of merging this. In https://github.com/xmonad/xmonad/pull/242 and https://github.com/xmonad/xmonad-contrib/pull/398 I started working on being able to query the current layout state. The next step is to add a `InspectLayoutModifier` type class to xmonad-contrib to allow inspecting the current state of some modifier. A wip version of that is sitting here: https://github.com/liskin/xmonad-contrib/commit/77c1821ed6e70668f02e0f5f28e11262209d0ded. This class will need instances for many layout combinators, including `Choose`, so I also need the constructors to be exported.

Can you please rebase and fix the conflicts?
@psibi, @byorgey can you folks then merge it please?
Thanks!

psibi:
@jumper149 Thanks for for the MR!

@liskin Done.

liskin:
This ended up breaking compilation of xmonad-contrib as modules that define their own L/R data constructors (X.U.Types, X.L.GridVariants, X.L.ResizeScreen, X.L.TallMastersCombo, X.C.Types) cause ambiguity now.
As LR was never exported before, the best course of action is probably to rename it and its constructors so it doesn't conflict. Exporting `L` and `R` from `XMonad` is really nasty. :-/

And then maybe also reduce the duplication in xmonad-contrib: use the renamed/exported LR in TallMastersCombo.

Reported here: https://old.reddit.com/r/xmonad/comments/ka1qxg/fixing_fullscreen/

PR #221, opened 2020-05-13
Use current screen to set dimensions of new floating windows

liskin:
### Description

This fixes a bug when using multiple screens with different dimensions,
causing some floating windows to be smaller/larger than the size they
requested.

Some applications (e.g. pinentry) always map their window at (0, 0) so
floatLocation would compute the window size relative to the screen
containing (0, 0) and if the current workspace is on another screen with
a different size, this relative size results in a different absolute
size, which is undesirable for fixed size floats.

Other applications like ssh-agent place their window at the center of
the framebuffer (ignoring xinerama layout). Same problem.

Then there are apps that remember their position/size when minimizing to
tray and then attempt to restore it when reopened. Again, if they
restore it on another screen, we miscalculate the size.

The fix is to use the current screen for calculating dimensions of new
(not yet mapped) floating windows.

Co-Authored-By: @Narthorn (https://github.com/xmonad/xmonad/pull/197)

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## unknown (unknown)
 
+  * Fixed a bug when using multiple screens with different dimensions,
+    causing some floating windows to be smaller/larger than the size they
+    requested.
+
 ## 0.15 (September 30, 2018)
 
   * Reimplement `sendMessage` to deal properly with windowset changes made
--- a/src/XMonad/Operations.hs
+++ b/src/XMonad/Operations.hs
@@ -24,6 +24,7 @@ import Data.Maybe
 import Data.Monoid          (Endo(..),Any(..))
 import Data.List            (nub, (\\), find)
 import Data.Bits            ((.|.), (.&.), complement, testBit)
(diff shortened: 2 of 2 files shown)


liskin:
This started here: https://github.com/xmonad/xmonad/pull/197
As I changed the code substantially, I'm now submitting it as my own pull request with a Co-Authored-By crediting Vincent.
I've been running the code for several months without any problems. What do the others think?

rajivr:
I am very new to Xmonad. I just integrated this patch into my Xmonad build and it seems to resolve the issue. Thanks a lot @liskin

malnajdi:
Is there a way for me to test this if i am installing Xmonad from APT ?

liskin:
@malnajdi Not really. You'll need to compile from git.

malnajdi:
> @malnajdi Not really. You'll need to compile from git.

Thanks.

wejn:
Thanks for the fix @liskin. This issue was plaguing me for a while, but I thought it's somehow my fault (a glitch in my setup).

PR #251, opened 2020-12-11
Add github CI integration

psibi:
### Description

Replaces Travis CI with github actions since travis doesn't work anymore: https://news.ycombinator.com/item?id=25338983

### Checklist

  - [ ] I've confirmed these changes don't belong in xmonad-contrib instead


psibi:
I will merge this in a couple of days unless there is objections. CC: @byorgey and others.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,39 @@
...
+
+jobs:
+  build:
+    name: CI
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest]
+        resolver: [lts-16, lts-15, lts-14]
```
liskin: We claim support all the way back to GHC 8.0 in xmonad.cabal, so we should probably test with lts-7 or 9 and lts-11 and lts-12 as well.

psibi: Sure, we can probably. I will update the resolvers section tomorrow to see if the CI passes. If it doesn't pass, I wouldn't be motivated to fix them though (via a separate stack.yaml file) and would likely remove them. :grinning:

liskin: Debian Stable has 8.4.4 so I'd probably attempt to support lts-12+ but anything older can probably be dropped?

Oh and EPEL8 has ghc 8.2.2. How likely is it someone's running EL8 on desktop and uses xmonad and needs new stuff to work?

psibi: > Debian Stable has 8.4.4 so I'd probably attempt to support lts-12+ but anything older can probably be dropped?

Yeah, I'm leaning towards the same. Sticking to support last three major version of ghc would likely make things simpler. I just added `lts-12` and the CI passes.

PR #252, opened 2020-12-13
Get rid of unused imports

slotThe:
### Description

Starting with  28e75da77fa1918ec3ef4ce63ada32507a75b50c we only support GHC versions 8.4.3 and up (more precisely, the GHC version associated with stackage lts-12.0 and up).  This is a good time to get rid of all the imports that are now actually unused for all supported versions.

This pr produces no unused import warnings up to GHC 8.6.x and one unused import warning for GHC 8.8.x (this is due to MFP being implemented in 8.8.x)

I already have an equivalent pr ready for `xmonad-contrib`.  However, it seems like [the resolver there](https://github.com/xmonad/xmonad-contrib/blob/master/.github/workflows/tests.yml#L17) only starts at `lts-14.0` instead of `lts-12.0`, which probably needs to be fixed first.  cc @psibi   

### Checklist

  - [X] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

### DIFF
--- a/xmonad.cabal
+++ b/xmonad.cabal
@@ -84,7 +84,11 @@ library
                  , setlocale
                  , unix
                  , utf8-string           >= 0.3 && < 1.1
-  ghc-options:     -funbox-strict-fields -Wall -fno-warn-unused-do-bind
+  ghc-options:     -funbox-strict-fields -Wall -fno-warn-unused-do-bind -Wno-unused-imports
+
+  -- Keep this in sync with the oldest version in 'tested-with'
+  if impl(ghc <= 8.4.4)
+     ghc-options: -Werror=unused-imports
 
   if flag(testing)
     buildable: False
@@ -92,7 +96,11 @@ library
 executable xmonad
   main-is:       Main.hs
   build-depends: base, X11, mtl, unix, xmonad
-  ghc-options:   -Wall -fno-warn-unused-do-bind
+  ghc-options:   -Wall -fno-warn-unused-do-bind -Wno-unused-imports
+
+  -- Keep this in sync with the oldest version in 'tested-with'
+  if impl(ghc <= 8.4.4)
+     ghc-options: -Werror=unused-imports
 
 executable generatemanpage
   main-is:        GenerateManpage.hs
--- a/xmonad.cabal
+++ b/xmonad.cabal
@@ -27,8 +27,7 @@ author:             Spencer Janssen, Don Stewart, Adam Vogt, David Roundy, Jason
                     Ondřej Súkup, Paul Hebble, Shachaf Ben-Kiki, Siim Põder, Tim McIver,
                     Trevor Elliott, Wouter Swierstra, Conrad Irwin, Tim Thelion
 maintainer:         xmonad@haskell.org
-tested-with:        GHC == 8.0.2, GHC == 8.2.2, GHC == 8.4.3, GHC == 8.6.5,
-                    GHC == 8.8.4, GHC == 8.10.2
+tested-with:        GHC == 8.4.4, GHC == 8.6.5, GHC == 8.8.3, GHC == 8.8.4
 category:           System
 homepage:           http://xmonad.org
 bug-reports:        https://github.com/xmonad/xmonad/issues
--- a/src/XMonad/Core.hs
+++ b/src/XMonad/Core.hs
@@ -36,7 +36,6 @@ import XMonad.StackSet hiding (modify)
 import Prelude
 import Control.Exception.Extensible (fromException, try, bracket, throw, finally, SomeException(..))
 import qualified Control.Exception.Extensible as E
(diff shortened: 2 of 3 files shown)


liskin:
If testing only with 8.4+ is desired, we should update the `tested-with:` field in xmonad.cabal.

I think that we should try to silence unused imports warnings for all ghc versions, and there are two ways to do that:
* use CPP and check version of base and other packages
* enable `-Wno-unused-imports`

I used to use the former a couple years back at work and it's a major pain, so I am strongly in favor of the latter. The downside is that it will allow the code to bitrot as all such warnings will be silenced forever. So how about this: add condition on ghc version to the cabal file — if ghc is the oldest we support, warn about unused imports, otherwise silence the warnings. This will ensure the code doesn't bitrot.

Anyway, all of this can be done in a followup, this PR can be merged as is. Thanks!

slotThe:
On Sun, Dec 13 2020 07:21, Tomáš Janoušek wrote:
> If testing only with 8.4+ is desired, we should update the
> `tested-with:` field in xmonad.cabal.
>
> I think that we should try to silence unused imports warnings for all
> ghc versions, and there are two ways to do that:
> * use CPP and check version of base and other packages
> * enable `-Wno-unused-imports`
>
> I used to use the former a couple years back at work and it's a major
> pain, so I am strongly in favor of the latter. The downside is that it
> will allow the code to bitrot as all such warnings will be silenced
> forever. So how about this: add condition on ghc version to the cabal
> file — if ghc is the oldest we support, warn about unused imports,
> otherwise silence the warnings. This will ensure the code doesn't
> bitrot.

Oh I really like these solutions actually; it kind of fits into this pr,
so I'll try to implement this here, if you don't mind.

liskin:
Sure, go ahead. :-)

slotThe:
@liskin is this about what you had in mind?

xmonad.cabal
```diff
@@ -84,15 +84,23 @@ library
                  , setlocale
                  , unix
                  , utf8-string           >= 0.3 && < 1.1
-  ghc-options:     -funbox-strict-fields -Wall -fno-warn-unused-do-bind
+  ghc-options:     -funbox-strict-fields -Wall -fno-warn-unused-do-bind -Wno-unused-imports
+
+  -- Keep this in sync with the oldest version in 'tested-with'
+  if impl(ghc == 8.4.3)
+     ghc-options: -Wunused-imports
```
liskin: The oldest version we actually test with is 8.4.4 (lts-12.26), so `ghc <= 8.4.4` is probably better.
I wonder if we want to `-Werror=unused-imports` or whether we're going to check the warnings manually.

liskin:
Yes, I love it, just a minor nitpick (as usual). :-)

xmonad.cabal
slotThe: Alright, I'll change that to `ghc <= 8.4.4`.  The `-Werror=unused-imports` bit sounds nice as well; a CI failure on one's pr is probably the best incentive possible for fixing these warnings, thanks!

liskin: Cool.

(I wasn't sure about the -Werror bit because it's incosistent with the other warnings. But once we have reliable CI we can enable those as well. :-))

xmonad.cabal
```diff
@@ -27,8 +27,7 @@ author:             Spencer Janssen, Don Stewart, Adam Vogt, David Roundy, Jason
                     Ondřej Súkup, Paul Hebble, Shachaf Ben-Kiki, Siim Põder, Tim McIver,
                     Trevor Elliott, Wouter Swierstra, Conrad Irwin, Tim Thelion
 maintainer:         xmonad@haskell.org
-tested-with:        GHC == 8.0.2, GHC == 8.2.2, GHC == 8.4.3, GHC == 8.6.5,
-                    GHC == 8.8.4, GHC == 8.10.2
+tested-with:        GHC == 8.4.3, GHC == 8.6.5, GHC == 8.8.4, GHC == 8.10.2
```
liskin: 8.4.4 here as well, perhaps?

liskin (approved):

xmonad.cabal
slotThe: Ah good point, I suppose we should only have versions there that are in the CI (I built it with 8.4.3 just now, so technically it's tested with that, but if the CI builds against 8.4.4 there's the possibility of this becoming outdated of course)

liskin: It's unlikely to matter, but the comment below says "Keep this in sync" so it'd be nice if it was kept in sync at least initially until someone inevitably forgets :-D

slotThe: Hah, very good point :) 

I'll reduce the version to only the ones in the CI then; `tested-with` doesn't mean "doesn't work with newer versions", after all

slotThe:
@psibi what do you think?

psibi (approved):

psibi:
Thanks!

PR #258, opened 2020-12-27
Revise XDG handling

### LINKED ISSUE #164: ~/.xmonad folder being created with configuration in XDG_CONFIG_HOME
### Problem Description

The xmonad.hs configuration is put in the xmonad folder within `XDG_CONFIG_HOME` (`~/.config/xmonad`), no `~/.xmonad` folder exists. Recompilation works correctly, but upon finishing a `~/.xmonad` folder exists, breaking a next recompile if the `.xmonad` is not being removed.

I was able to resolve this, by manually creating the xmonad folder in the `XDG_DATA_HOME` one.

The fact that all xmonad folders in each of the XDG ones needed to be created was not obvious from the documentation (changelog). I don't know whether this is expected or not, so not sure what the action points are (if any). I would definitely suggest making this more obvious in the documentation, and maybe also have xmonad create those when the `~/.xmonad` folder doesn't exist, and the configuration file is detected in the `XDG_CONFIG_HOME/xmonad` folder (?)

pinging @geekosaur as I discussed my observation (without finding the cause at that point) on `#xmonad`

slotThe:
### Description

Fixes #164 and #256

I'm putting this out here in a relatively early state, because I know
opinions are to be had on a lot of these things, so I'd rather not wait
until this is more mature for people to say I've done something horribly
wrong.  

As per #256, this puts forth the following changes to how we handle the
whole config situation.

  1. If all three of xmonad's environment variables (`XMONAD_DATA_DIR,`
     `XMONAD_CONFIG_DIR`, and `XMONAD_CACHE_DIR`) are set, use them.
  2. If there is a build script `build` or configuration `xmonad.hs` in
     `~/.xmonad`, set all three directories to `~/.xmonad`.
  3. Otherwise, use `XDG_DATA_HOME`, `XDG_CONFIG_HOME`, and
     `XDG_CACHE_HOME` (or their respective fallbacks).

If none of the above exist, we default to using the XDG_* variables,
creating the necessary directories if needed.

In the cases of 1. and 3., the build script or executable is expected to
be in the config dir (this is the case at the moment as well).

This doesn't quite compile with `xmonad-contrib` yet, though there
aren't very many changes needed to make it work.  I've created a quick
fix [on this branch](https://github.com/slotThe/xmonad-contrib/tree/xdg-mess), in case anyone wants to try this.  From my (very) light
testing so far, it at least seems to do what it's supposed to.

### Checklist


slotThe:
@geekosaur @liskin @psibi what do you people think?  

I think the abstract approach mentioned in #256 is sound and this implementation "works for me" and shouldn't break any existing configs

CHANGES.md
```diff
@@ -15,6 +15,19 @@
 
   * Compatibility with GHC 9.0
 
+  * Improve handling of XDG directories.
+
+      1. If all three of xmonad's environment variables (`XMONAD_DATA_DIR,`
+         `XMONAD_CONFIG_DIR`, and `XMONAD_CACHE_DIR`) are set, use them.
+      2. If there is a build script `build` or configuration `xmonad.hs` in
```
psibi: What does build script `build` mean ? I think you are referring to a file named `~/build`. Probably clarifying that it would be helpful for newcomers.

src/XMonad/Core.hs
```diff
@@ -447,126 +449,94 @@ runOnWorkspaces job = do
...
+    , cacheDir :: !FilePath
+    }
+    deriving (Show)
+
+-- | Build up the 'Dirs' that xmonad will use.  They are chosen as
+-- follows:
+--
+-- 1. If all three of xmonad's environment variables (@XMONAD_DATA_DIR@,
+--    @XMONAD_CONFIG_DIR@, and @XMONAD_CACHE_DIR@) are set, use them.
+-- 2. If there is a build script @build@ or configuration @xmonad.hs@ in
```
psibi: Same comment as above.

psibi (approved):
LGTM apart from minor haddock issue. I will merge this in a couple of days unless there is objection from others.

CHANGES.md
slotThe: I'm referring to something like `~/.xmonad/build` or `~/.config/xmonad/build`; there are [examples](https://github.com/xmonad/xmonad-testing/tree/master/build-scripts) for this in the `xmonad-testing` repository. 

I'll try to improve the docs a bit there

src/XMonad/Core.hs
```diff
@@ -447,126 +449,95 @@ runOnWorkspaces job = do
...
+-- 1. If all three of xmonad's environment variables (@XMONAD_DATA_DIR@,
+--    @XMONAD_CONFIG_DIR@, and @XMONAD_CACHE_DIR@) are set, use them.
+-- 2. If there is a build script called @build@ or configuration
+--    @xmonad.hs@ in @~\/.xmonad@, set all three directories to
+--    @~\/.xmonad@.
+-- 3. Otherwise, use @XDG_DATA_HOME@, @XDG_CONFIG_HOME@, and
+--    @XDG_CACHE_HOME@ (or their respective fallbacks).
+--
+-- If none of the above exist, then default to using the @XDG_*@
+-- variables, creating the necessary directories if needed.
```
liskin: This doesn't seem to add much to point 3, especially since the variables don't need to be set. Might want to drop this and move the "creating the necessary directories if needed" to point 3.

src/XMonad/Core.hs
```diff
@@ -447,126 +449,95 @@ runOnWorkspaces job = do
...
+-- The xmonad configuration file (or the build script, if present) is
+-- always assumed to be in @cfgDir@.
+--
+getDirs :: IO Dirs
+getDirs = do
+    let xmEnvs  = ["XMONAD_CONFIG_DIR", "XMONAD_DATA_DIR", "XMONAD_CACHE_DIR"]
+        xdgEnvs = [XdgData, XdgConfig, XdgCache]
+
+    -- Check for xmonads' environment variables first
+    toDirs . sequenceA <$> traverse getEnv xmEnvs >>= \case
```
liskin: Just a nitpick: I find this hard to read as the order of execution/evaluation is 1. traverse, 2. toDirs, 3. case. When `.` and `fmap` are used, I prefer `=<<` to keep it right-to-left, and when `case` is needed, `x <- …` and `case x of` on another line are easier to understand.

src/XMonad/Core.hs
```diff
@@ -447,126 +449,95 @@ runOnWorkspaces job = do
...
+--
+-- If none of the above exist, then default to using the @XDG_*@
+-- variables, creating the necessary directories if needed.
+--
+-- The xmonad configuration file (or the build script, if present) is
+-- always assumed to be in @cfgDir@.
+--
+getDirs :: IO Dirs
+getDirs = do
+    let xmEnvs  = ["XMONAD_CONFIG_DIR", "XMONAD_DATA_DIR", "XMONAD_CACHE_DIR"]
```
liskin: The order here is wrong.

I tried to make the code more _obviously correct_, and I went perhaps too far with that, so I'll just post the result and let you choose whether to get inspired or not. :-)

```haskell
getDirs :: IO (Dirs FilePath)
getDirs = xmEnvDirs <|> xmDirs <|> xdgDirs
  where
    xmEnvDirs = do
        -- Check for xmonad's environment variables first
        let xmEnvs = Dirs{ dataDir = "XMONAD_DATA_DIR", cfgDir = "XMONAD_CONFIG_DIR", cacheDir = "XMONAD_CACHE_DIR" }
        maybe empty pure =<< sequenceA <$> traverse getEnv xmEnvs
    xmDirs = do
        -- Check whether the config file or a build script is in the
        -- @~\/.xmonad@ directory
        xmDir <- getAppUserDataDirectory "xmonad"
        conf <- doesFileExist $ xmDir </> "xmonad.hs"
        build <- doesFileExist $ xmDir </> "build"
        guard $ conf || build
        -- Place everything there if yes
        pure Dirs{ dataDir = xmDir, cfgDir = xmDir, cacheDir = xmDir }
    xdgDirs = for Dirs{ dataDir = XdgData, cfgDir = XdgConfig, cacheDir = XdgCache } $ \d -> do
        -- Use XDG directories as a fallback
        d' <- getXdgDirectory d "xmonad"
        createDirectoryIfMissing True d'
        pure d'
```

src/XMonad/Core.hs
liskin: (possibly irrelevant if you choose to adopt the refactored version from the comment above)

src/XMonad/Core.hs
liskin: (this applies to the CHANGES.md entry as well)

liskin (changes requested):
Good work fixing this stuff! :+1: 

Found a mistake in env var handling, the rest are just doc/code improvement suggestions.

src/XMonad/Core.hs
slotThe: I'm always for using `IO`s `Alternative` instance when it's safe to do so, so big yes (and thanks!) from me :)

src/XMonad/Core.hs
```diff
@@ -447,126 +449,97 @@ runOnWorkspaces job = do
...
+-- such as the run-time state file and the configuration binary
+-- generated by GHC.
+--
+-- * @cfgDir@: This directory is where user configuration files are
+-- stored (e.g, the xmonad.hs file).  You may also create a @lib@
+-- subdirectory in the configuration directory and the default recompile
+-- command will add it to the GHC include path.
+--
+-- * @cacheDir@: This directory is used to store temporary files that
+-- can easily be recreated.  For example, the XPrompt history file.
```
liskin, line 465: It's a bit funny that the configuration binary, which can be easily recreated, gets placed in dataDir, whereas the prompt history, which requires superhuman memory abilities or superhuman backup regime to recreate, goes in cacheDir. But xmonad isn't the only program that does it this way so I guess it doesn't matter.

CHANGES.md
```diff
@@ -15,6 +15,21 @@
...
+      1. If all three of xmonad's environment variables (`XMONAD_DATA_DIR,`
+         `XMONAD_CONFIG_DIR`, and `XMONAD_CACHE_DIR`) are set, use them.
+      2. If there is a build script called `build` (see [here](https://github.com/xmonad/xmonad-testing/tree/master/build-scripts) for usage
+         examples) or configuration `xmonad.hs` in `~/.xmonad`, set all three
+         directories to `~/.xmonad`.
+      3. Otherwise, use the `xmonad` directory in `XDG_DATA_HOME`,
+         `XDG_CONFIG_HOME`, and `XDG_CACHE_HOME` (or their respective
+         fallbacks). These directories are created if necessary.
+
+    If none of the above exist, we default to using the `XDG_*` variables,
```
liskin: This sentence can be dropped as well I think.

liskin (approved):

PR #260, opened 2021-01-09
GenerateManpage: Port to cmdline pandoc and update manpage in CI

liskin:
We don't need bleeding edge pandoc and we don't need to use the Haskell library interface either. The command-line interface is more stable, so using that instead should reduce maintenance burden and simplify packaging (e.g. Debian patches GenerateManpage to work with the version of pandoc that's in Debian).

To reduce maintenance burden even further, invoke GenerateManpage in CI after push to master and, if changed, commit and push the result. Example CI run: https://github.com/liskin/xmonad/runs/1671786679?check_suite_focus=true

Also, reintroduce putting xmonad version into the manpage.


peti (changes requested):
This PR solves a problem that does not exist. We've needed almost no maintenance effort to keep this little tool running and I am against throwing away a solution that's working just fine for no apparent reason. 👎

liskin:
> This PR solves a problem that does not exist.

Not true. It solves several:

* xmonad doesn't build with +generatemanpage in LTS 16
* nor does it build in LTS 15, 14 or 12
* doesn't build with Stackage Nightly either
* therefore that code isn't being tested in our CI and the manpage isn't automatically kept up to date
* furthermore, it doesn't build in Debian unstable, as they only have pandoc 2.9.2.1, and therefore they need to [carry a patch](https://salsa.debian.org/haskell-team/DHG_packages/-/blob/master/p/xmonad/debian/patches/newer-pandoc) to fix the build (which they'll need to update again for the next xmonad release)
* doesn't build in Fedora either, they also have 2.9.2.1, but the Fedora package doesn't rebuild the manpage so it't not a problem

> We've needed almost no maintenance effort to keep this little tool running and I am against throwing away a solution that's working just fine for no apparent reason.

I believe getting the build to work with the stack-based builds that we currently have is itself desirable, even if the maintenance effort is otherwise little. I also believe that using a version of pandoc that few of the distros have so that they bear the maintenance effort of porting to other versions is quite selfish. Saying "working just fine" when it's pretty clear there are several widely used environments where it demonstrably doesn't work is the height of selfishness, in my book. :-/

That being said, I'm not entirely sure my solution is the best one. I feel like invoking pandoc from $PATH is a bit wrong. Maybe it'd be better for GenerateManpage to only output the markdown on stdout and do the rest in Makefile or something. That's up for debate, of course. But I really do believe there is a problem that is worth solving.

peti:
> It solves several:
> 
>     * xmonad doesn't build with +generatemanpage in LTS 16 
>     * nor does it build in LTS 15, 14 or 12
>     * doesn't build with Stackage Nightly either

The program compiles just fine:

~~~
$ cabal build generatemanpage
Resolving dependencies...
Build profile: -w ghc-8.10.3 -O1
In order, the following will be built (use -v for more details):
 - xmonad-0.15 (exe:generatemanpage) (first run)
Configuring executable 'generatemanpage' for xmonad-0.15..
Preprocessing executable 'generatemanpage' for xmonad-0.15..
Building executable 'generatemanpage' for xmonad-0.15..
[1 of 1] Compiling Main             ( util/GenerateManpage.hs, /home/simons/src/xmonad/dist-newstyle/build/x86_64-linux/ghc-8.10.3/xmonad-0.15/x/generatemanpage/build/generatemanpage/generatemanpage-tmp/Main.o )
Linking /home/simons/src/xmonad/dist-newstyle/build/x86_64-linux/ghc-8.10.3/xmonad-0.15/x/generatemanpage/build/generatemanpage/generatemanpage ...
~~~

Not everyone uses `stack`.

> that code isn't being tested in our CI and the manpage isn't automatically kept up to date

Yeah, apparently someone had the bright idea to base our CI builds on the most rigid and conservative build tool there is because surely we wouldn't want to notice in CI when a new version of our dependencies comes out that introduces an incompatible change, would we? Anyhow, https://github.com/xmonad/xmonad/pull/261 fixes that issue.

> it doesn't build in Debian unstable, as they only have pandoc 2.9.2.1

`generatemanpage` is a tool used by *developers*. I don't see why Debian would want to run it in the first place. We include the generated man page in our release archive. Distributions that re-build that file anyway mostly waste energy, increase CO2 emissions, and ensure that mankind's extinction happens a little earlier than it would otherwise. I don't see why we should bend over backwards to accommodate that workflow.

> I believe getting the build to work with the stack-based builds that we currently have is itself desirable, even if the maintenance effort is otherwise little.

Sure, I am totally in favor of making sure that stack can build `generatemanpage`. I would prefer to fix the Stack build over completely re-writing our tool, though. The current approach is extremely reliable, declarative, re-producible, and transparent. All dependencies are listed in the Cabal file and any reasonable build tool can build them. The approach that relies on `stack` being magically in `$PATH` for some reason, however, breaks all those nice properties.

> Saying "working just fine" when it's pretty clear there are several widely used environments where it demonstrably doesn't work is the height of selfishness, in my book. :-/

Would you mind cutting out the *ad hominem* attacks, please? I am perfectly happy to have a wild, controversial discussion that's based on fact and rational thought, but I won't waste my time talking to people who pass value judgement on my character because I don't agree with their PR.

liskin:
On Sat, Jan 09, 2021 at 09:34:13AM -0800, Peter Simons wrote:
>> Saying "working just fine" when it's pretty clear there are several widely used environments where it demonstrably doesn't work is the height of selfishness, in my book. :-/
>
>Would you mind cutting out the *ad hominem* attacks, please? I am perfectly happy to have a wild, controversial discussion that's based on fact and rational thought, but I won't waste my time talking to people who pass value judgement on my character because I don't agree with their PR.

Okay, I'm sorry for overreacting. It's still worth noting that I'm not 
angry because you disagree with my PR; I'm angry because you dismissed 
the premise of the PR as being a nonexistent problem, and I strongly 
disagree. If it was not a problem, Debian wouldn't be patching over it, 
and I'd be able to build and run generatemanpage without noticing any 
trouble.

Let's keep the discussion civilized then.

>The program compiles just fine:
>
>~~~
>[...]
>~~~
>
>Not everyone uses `stack`.

Exactly. Not everyone uses `cabal` either. Debian package builds don't 
use any of the two, in fact, they package the build deps and install 
them using apt/dpkg. All of these methods are, IMO, valid and should be 
supported.

You've got a point that perhaps they don't need to use generatemanpage 
at all, in which case we could ignore that workflow and only support 
`cabal` and `stack`. But I think maybe Debian has a policy that says 
everything that can be rebuilt should be rebuilt or something? Anyway I 
wouldn't be comfortable telling them not to rebuild without having the 
rebuild in CI. I don't feel like manually checking that PR authors 
rebuilt the manpage correctly.

>Yeah, apparently someone had the bright idea to base our CI builds on the most rigid and conservative build tool there is because surely we wouldn't want to notice in CI when a new version of our dependencies comes out that introduces an incompatible change, would we? Anyhow, https://github.com/xmonad/xmonad/pull/261 fixes that issue.

That someone was fairly clear that it's a minimal viable solution and 
`cabal` CI builds should be added as well sooner or later and asked for 
help with that. I don't think they deserve this blame.

Testing with latest versions of deps is definitely something that's 
missing, and we should add that. I don't expect much breakage, though 
(we still use `Control.Exception.Extensible` and everything's fine, …), 
so it's a low priority task for me. Not being able to build 
`generatemanpage` is a more pressing issue from my point of view.

>`generatemanpage` is a tool used by *developers*. I don't see why Debian would want to run it in the first place. We include the generated man page in our release archive. Distributions that re-build that file anyway mostly waste energy, increase CO2 emissions, and ensure that mankind's extinction happens a little earlier that it would otherwise. I don't see why we should bend over backwards to accommodate that workflow.

As I said above, I think there may be several valid reasons to rebuild 
the man page. And I definitely don't think the tool should be run by 
developers personally: it should either run in CI or during build, or 
both. But if it was a tool to be run by devs personally, that would be 
an argument **for** supporting older pandoc versions, not against. :-)

(The CO₂ argument seems weird in this context. You can either choose to 
save the environment and use possibly not bleeding-edge distro-packaged 
binaries or prioritize having reliable reproducible dependencies handled 
by a Haskell-specific build tool. I don't think you can have both. But 
please let's not argue about this.)

>Sure, I am totally in favor of making sure that stack can build `generatemanpage`. I would prefer to fix the Stack build over completely re-writing our tool, though. The current approach is extremely reliable, declarative, re-producible, and transparent. All dependencies are listed in the Cabal file and any reasonable build tool can build them. The approach that relies on `stack` being magically in `$PATH` for some reason, however, breaks all those nice properties.

I agree that relying on `pandoc` being in path feels weird. On the other 
hand, I feel that we really should support pandoc versions going back to 
at least 2.9, and if that's going to involve CPP in GenerateManpage.hs, 
I think that's even worse. :-/

(I do have to admit I'm not a big fan of language-specific build tools, 
though. They do a terrible job of cleaning up after themselves, compared 
to tools like apt or rpm, and there are other downsides as well. That 
may explain why using `pandoc` binary feels only slightly more weird 
than using something like `sed`, and why building `pandoc` using 
`cabal`/`stack` is something I look critically upon.)

I mean, GenerateManpage.hs could probably even be a oneliner in 
sed/perl/awk or something. It's hard to argue that pulling several 
hundred megs of deps and building them is really worth it, or that it 
saves the environment, or anything. The only argument here that I 
acknowledge is that it's worked for years. And that's a good argument, 
don't get me wrong, but I really do feel like it's not working well if I 
can't run it and Debian can't run it.

psibi:
@peti 

 > Yeah, apparently someone had the bright idea to base our CI builds on the most rigid and conservative build tool there is because surely we wouldn't want to notice in CI 
 
 That someone is me. The travis CI got broken and none of the xmonad projects had any CI running, so I integrated stack based CI as that's the one I use. Cabal based CI patches are welcome.

peti:
@psibi,

> The travis CI got broken and none of the xmonad projects had any CI running, so I integrated stack based CI as that's the one I use.

Do you recall what exactly broke? I remember generating the travis-ci.yml file with the Haskell tool `haskell-ci`, which creates a rather sophisticated build and I never realized that it stopped working? I only vaguely recall seeing a PR that switched to Github actions a while ago, but I didn't pay attention because I kinda assumed that there wouldn't be a switch unless the new Ci has feature parity with the old one, but apparently the circumstances where different than I thought they'd be.

geekosaur:
Travis went away.

peti:
I'm using Travis right now for two dozen different projects. My understanding is that it will go away soon, kinda, but it hasn't yet.

psibi:
> Do you recall what exactly broke? I remember generating the travis-ci.yml file with the Haskell tool haskell-ci, which creates a rather sophisticated build and I never realized that it stopped working?

The travis CI didn't run any of the new builds. Other open source projects were facing the same issue at the time.

psibi (approved):

liskin:
Replaced by #285 or #286.

PR #261, opened 2021-01-09
stack.yaml: switch from Stackage LTS to Nightly

peti:
Let's use a moderately recent compiler like ghc-8.10.3
instead of the old 8.8.x branch for development. Same goes
for our dependencies; development should take place with
recent versions of our dependencies.

This PR also fixes the build of `generatemanpage` and enables
it in our CI environment.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        resolver: [lts-16, lts-15, lts-14, lts-12]
+        resolver: [nightly]
 
     steps:
       - name: Clone project
--- a/stack.yaml
+++ b/stack.yaml
@@ -1,7 +1,24 @@
-resolver: lts-16.22
+resolver: nightly-2021-01-05
+
+flags:
+  xmonad:
+    generatemanpage: true
 
 packages:
 - ./
 
 extra-deps:
 - X11-1.9.2
+- pandoc-2.11.3.2
+- citeproc-0.3.0.3
+- commonmark-0.1.1.2
+- commonmark-extensions-0.2.0.4
+- commonmark-pandoc-0.2.0.1
+
+nix:
+  packages:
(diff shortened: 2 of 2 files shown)


liskin:
Doesn't seem like it actually fixes anything, does it?

And it doesn't help the Debian folks one bit either. :-(

peti:
> Doesn't seem like it actually fixes anything, does it?

Why? Have you tried running `stack build` with that PR applied? Did it not work?

liskin:
> Why? Have you tried running `stack build` with that PR applied? Did it not work?

I haven't, personally, but GitHub Actions have.

peti:
You should really try it yourself. You might learn something.

peti:
Now CI builds fine, too.

slotThe:
@peti Why was this merged?  As far as I can tell, no consensus was reached regarding which approach (this or #260) is to be favoured.  

Only building with stack nightly is highly problematic in its own right—we, for example, have [this](https://github.com/xmonad/xmonad/blob/master/xmonad.cabal#L30) in our cabal file, promising to support GHC versions `8.4.4` and up; not just `8.10.x`.  Also [this](https://github.com/xmonad/xmonad/blob/master/xmonad.cabal#L90) way of detecting unused imports is now not checked in the CI; does it now fall onto the developer to check this manually?    

I'm voting to revert this until a proper consensus has been reached.

liskin:
@peti Seriously? This fixes none of the problems and breaks most of the testing.

@byorgey @geekosaur Can you guys please step in?

peti:
> @peti Why was this merged?

This was merged, because as a release manager for xmonad I want to be sure that the tool to generate our man page works, and the previous solution did not build it. I also want to make sure that we support *recent* compilers, whereas the previous solution focussed on old ones.

Having said that, I am perfectly happy to re-enable those old LTS builds asap. Please refer to https://github.com/xmonad/xmonad/pull/261/commits/3015968ee43a71648c260857d73dba19168083fe for details.

PR #263, opened 2021-01-13
Control.Exception.Extensible -> Control.Exception

slotThe:
### Description

According to its documentation[1], this module simply re-exports
Control.Exception on recent GHC versions.  As we only support recent
versions, this import is unnecessary.

[1] http://hackage.haskell.org/package/extensible-exceptions-0.1.1.4/docs/Control-Exception-Extensible.html

This should probably be merged on conjunction with xmonad/xmonad-contrib/pull/444

### Checklist

  ~~- [ ] I updated the `CHANGES.md` file~~
  Not needed as we've already dropped support for GHC versions that old.

### DIFF
--- a/src/XMonad/Core.hs
+++ b/src/XMonad/Core.hs
@@ -35,8 +35,8 @@ module XMonad.Core (
 import XMonad.StackSet hiding (modify)
 
 import Prelude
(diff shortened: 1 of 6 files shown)


liskin:
Would be nice if we had some CI in place to make sure this works in all the environments we expect people to actually use xmonad… Unfortunately someone had the bright idea to break that CI. :-/

PR #268, opened 2021-01-20
This enables testing with stackage resolvers

psibi:


stack.yaml
```diff
@@ -1,19 +1,11 @@
-resolver: nightly-2021-01-05
+resolver: lts-16.22
 
-flags:
```
peti, line 3: Please don't drop these flags from the `stack.yaml` file. It is important that this flag is enabled because without it continuous builds will not include the `genmanpage` executable.

psibi:
@peti Can I know what objections you have for this MR ?

The man page generation will be resolved by this MR: https://github.com/xmonad/xmonad/pull/260

.github/workflows/tests.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        resolver: [nightly]
```
peti, line 17: Please don't drop Stackage Nightly from the list of supported build environments. It is important that we know our code builds in the Nightly snapshot because that environment is going to turn into the next LTS release, eventually.

peti (changes requested):

stack.yaml
psibi: @peti This is going to be addressed by this MR: https://github.com/xmonad/xmonad/pull/260

peti:
> The man page generation will be resolved by this MR: #260

I disagree with that change for the reasons I've outlined in great detail in the PR.

psibi:
> I disagree with that change for the reasons I've outlined in great detail in the PR.

I see that you haven't replied to this response from @liskin there: https://github.com/xmonad/xmonad/pull/260#issuecomment-757349677

Also, I agree with the changes there and my intention is to get both of these merged.

.github/workflows/tests.yml
liskin: This is a good point and I'm sorry for not noticing this in #266. I think it could've been addressed in a followup commit and didn't need a revert, though.

psibi: My plan to include new lts as it's created. But that's a good point and I will update the MR to include that too.

peti:
> I see that you haven't replied to this response from @liskin there: [#260 (comment)](https://github.com/xmonad/xmonad/pull/260#issuecomment-757349677)

No, I haven't. And I won't because that response doesn't address the points I have made in my original response. I have better things to do than to repeat myself over and over talking to someone who doesn't want to hear what I'm saying.

> Also, I agree with the changes there and my intention is to get both of these merged.

Well, good for you.

liskin:
> I have better things to do than to repeat myself over and over talking to someone who doesn't want to hear what I'm saying.

This feeling is, too, mutual. :-/

stack.yaml
peti: #260 sucks and I don't want that change in our code. There are various reasons to reject that change, but the most important one is that it drops knowledge from our Cabal file about an explicit dependency on `pandoc` and turns that into a hidden dependency on `pandoc` being `$PATH` for some reason. Since our build tools don't know about that requirement, it means that builds will fail. In it's current form, the change is certainly going to break the Nix and NixOS builds, for example. It might also break the openSUSE Linux build, but I'm haven't actually checked the latest spec file there so I'm not entirely sure about that. Anyhow, implicit dependency *suck* and introducing them into our builds is the wrong direction to go.

liskin: We already implicitly depend on libx11-dev etc. so that argument is not as good as you think it is.
Additionally, it was you who said that generatemanpage is meant to be run by devs only, so the implicit dependency might not be a problem for distros at all, with the possible exception of Debian which may be needing to rebuild the manpage for policy reasons.

liskin: Also, breaking the package build isn't the end of the world. For nix, it's probably trivially resolved by adding pandoc to the list of nix packages in stack.yaml, and distros like Debian will be more than happy to switch from patching xmonad to having one additional build-dep.

peti: > We already implicitly depend on libx11-dev etc. 

We don't depend on `libx11-dev` at all. `xmonad` depends on the Haskell library `X11`, which in turn depends on a couple of system libraries that are *explicitly* declared as dependencies in it's Cabal file at https://github.com/xmonad/X11/blob/master/X11.cabal#L74. The change you proposed drops a direct dependency of `genmanpage` from our Cabal file. 

> Additionally, it was you who said that generatemanpage is meant to be run by devs only, so the implicit dependency might not be a problem for distros at all, with the possible exception of Debian which may be needing to rebuild the manpage for policy reasons.

You realize that you actually don't know that, right? You keep making this claim that Debian "needs" to rebuild the man page, but as a matter fact you have no freaking clue whether that's actually true or not.

peti: > Also, breaking the package build isn't the end of the world.

Oh? It appears that you worry about our support for Nix far less than you worry about our support for Debian. Interesting.

> For nix, it's probably trivially resolved by adding pandoc to the list of nix packages in stack.yaml.

No, that won't help a bit. Trust me on this. I know a thing or two about how xmonad is packaged in Nix.

peti: @psibi, why did you mark this conversation as resolved even though it clearly isn't?

psibi: I thought my second comment actually resolved it, but it supposedly didn't. :-)

liskin: > We don't depend on `libx11-dev` at all. `xmonad` depends on the Haskell library `X11`, which in turn depends on a couple of system libraries that are _explicitly_ declared as dependencies in it's Cabal file at https://github.com/xmonad/X11/blob/master/X11.cabal#L74. The change you proposed drops a direct dependency of `genmanpage` from our Cabal file.

Oh, mea culpa. Unfortunately in practice the explicit extra-libraries deps are mostly meaningless, as their semantics is undefined (what does X11 mean? why not Xlib, or xlib, or libx11-dev, or xlib-devel, or …?) and therefore distros don't use it. Maybe nix does, I don't know, but the other don't.

So, um, maybe adding pandoc to extra-libraries would be an option? :-)

> You realize that you actually don't know that, right?

Yes, indeed I do. I'm not the one rushing any merges (hint: you are). I'm waiting to get some replies on the mailing list thread.

> Oh? It appears that you worry about our support for Nix far less than you worry about our support for Debian. Interesting.

For fuck's sake, Peter, stop accusing me of your own wrongdoings!

I care about both, and I'm more than willing to look for an acceptable compromise, which is why I solicited feedback from the Debian packagers and asked for PR feedback before merging the PR. You, on the other hand, quite explicitly said that Debian's packaging is not an issue for you in the slightest, and you keep pushing commits that only consider your own environment.

> No, that won't help a bit. Trust me on this. I know a thing or two about how xmonad is packaged in Nix.

Okay, fair enough. So what needs to be done to have pandoc in scope when building xmonad for nix? And is it really that much of an issue? Why does Nix need to build generatemanpage and Debian not? Why do you believe it's okay for Debian to patch xmonad, when at the same time you claim that adding a single dependency to the Nix build is a total no-go?

Are you, by any chance, worrying about our support for Debian far less than you worry about our support for Nix?

peti (commented):
I'd like to change my previous review into a neutral one. I'm happy with whatever the outcome of this process is going to be.

liskin:
Done in https://github.com/liskin/xmonad/commit/fb390fa9ccb7a4c73a6b29d609bf30daf5e7d682

PR #271, opened 2021-02-02
Added Lenses for StackSet

auscyber:
### Description

This is a light series of lenses for the StackSet and associated datatypes. This would be useful in easily creating modules that edit the StackSet.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## unknown
 
+### Enhancements
+
+  * Added custom cursor shapes for resizing and moving windows.
+
 ## 0.17.0 (October 27, 2021)
 
 ### Enhancements
--- a/src/XMonad/Operations.hs
+++ b/src/XMonad/Operations.hs
@@ -654,14 +654,20 @@ float w = do
 
 -- | Accumulate mouse motion events
 mouseDrag :: (Position -> Position -> X ()) -> X () -> X ()
-mouseDrag f done = do
+mouseDrag = mouseDragCursor Nothing
+
+-- | Like 'mouseDrag', but with the ability to specify a custom cursor
+-- shape.
+mouseDragCursor :: Maybe Glyph -> (Position -> Position -> X ()) -> X () -> X ()
+mouseDragCursor cursorGlyph f done = do
     drag <- gets dragging
     case drag of
         Just _ -> return () -- error case? we're already dragging
         Nothing -> do
             XConf { theRoot = root, display = d } <- ask
-            io $ grabPointer d root False (buttonReleaseMask .|. pointerMotionMask)
-                    grabModeAsync grabModeAsync none none currentTime
+            io $ do cursor <- maybe (pure none) (createFontCursor d) cursorGlyph
+                    grabPointer d root False (buttonReleaseMask .|. pointerMotionMask)
+                      grabModeAsync grabModeAsync none cursor currentTime
             modify $ \s -> s { dragging = Just (motion, cleanup) }
  where
     cleanup = do
@@ -679,7 +685,9 @@ mouseMoveWindow w = whenX (isClient w) $ withDisplay $ \d -> do
     (_, _, _, ox', oy', _, _, _) <- io $ queryPointer d w
     let ox = fromIntegral ox'
         oy = fromIntegral oy'
-    mouseDrag (\ex ey -> do
+    mouseDragCursor
+              (Just xC_fleur)
+              (\ex ey -> do
                   io $ moveWindow d w (fromIntegral (fromIntegral (wa_x wa) + (ex - ox)))
                                       (fromIntegral (fromIntegral (wa_y wa) + (ey - oy)))
                   float w
@@ -692,12 +700,13 @@ mouseResizeWindow w = whenX (isClient w) $ withDisplay $ \d -> do
     wa <- io $ getWindowAttributes d w
     sh <- io $ getWMNormalHints d w
     io $ warpPointer d none w 0 0 0 0 (fromIntegral (wa_width wa)) (fromIntegral (wa_height wa))
-    mouseDrag (\ex ey -> do
+    mouseDragCursor
+              (Just xC_bottom_right_corner)
+              (\ex ey -> do
                  io $ resizeWindow d w `uncurry`
                     applySizeHintsContents sh (ex - fromIntegral (wa_x wa),
                                                ey - fromIntegral (wa_y wa))
                  float w)
-
               (float w)
 
 -- ---------------------------------------------------------------------


geekosaur:
I'm not sure we want a dependency on any of the lens packages. Perhaps it's
time for xmonad-lens, since I'm also not sure it belongs in xmonad-extras.

auscyber:
the microlens package is quite small. like no external dependencies

auscyber:
I also manually wrote all the lenses. so no template Haskell either. i could probably even remove the aliases. but they would just be usable in things anyway

slotThe:
On Tue, Feb 02 2021 04:44, brandon s allbery kf8nh wrote:
> I'm not sure we want a dependency on any of the lens packages. Perhaps it's
> time for xmonad-lens, since I'm also not sure it belongs in xmonad-extras.

I would agree that his is perhaps not a good fit for xmonad core.  It
may or may not be in scope for xmonad-extras; @psibi (since you've been
in charge of that repo for the past few years) what do you think?

The dependency in this case is doubly weird, because we're only using
the `Lens'` type from microlens, which may well be directly defined as

    type Lens' s a = forall f. Functor f => (a -> f a) -> s -> f s

in that module.

auscyber:
As I said. It could be removed and just be the actual functions just labelled as lenses. Without the type aliases

psibi:
>  It may or may not be in scope for xmonad-extras; @psibi (since you've been
in charge of that repo for the past few years) what do you think?

I don't have strong opinions, but I'm leaning towards having it as part of separate package itself.

ghost:
There's a lot of stuff that these kinds of optics clean up. I've done a similar thing in https://github.com/wygulmage/xmonad/tree/optics_1_setters
All the functions like `mapLayout`, `mapWorkspaces`, `screens`, `workspaces`, etc. become trivial.

microlens is fantastic, but all you need defined to make them useful are `%%~`, `%~` and `^.`.

For example, in `XMonad.Core`, `runOnWorkspaces` goes from this:
```
  runOnWorkspaces job = do
    ws <- gets windowset
    h <- mapM job $ hidden ws
    c:v <- mapM (\s -> (\w -> s { workspace = w}) <$> job (workspace s))
             $ current ws : visible ws
    modify $ \s -> s { windowset = ws { current = c, visible = v, hidden = h } }
```
to this:
```
  runOnWorkspaces job = get >>= (windowsetLens . workspacesTraversal %%~ job) >>= put
```

I hope that's a good argument for putting it as close to the core as possible.

auscyber:
> There's a lot of stuff that these kinds of optics clean up. I've done a similar thing in https://github.com/wygulmage/xmonad/tree/optics_1_setters
> All the functions like `mapLayout`, `mapWorkspaces`, `screens`, `workspaces`, etc. become trivial.
> 
> microlens is fantastic, but all you need defined to make them useful are `%%~`, `%~` and `^.`.
> 
> For example, in `XMonad.Core`, `runOnWorkspaces` goes from this:
> 
> ```
>   runOnWorkspaces job = do
>     ws <- gets windowset
>     h <- mapM job $ hidden ws
>     c:v <- mapM (\s -> (\w -> s { workspace = w}) <$> job (workspace s))
>              $ current ws : visible ws
>     modify $ \s -> s { windowset = ws { current = c, visible = v, hidden = h } }
> ```
> 
> to this:
> 
> ```
>   runOnWorkspaces job = get >>= (windowsetLens . workspacesTraversal %%~ job) >>= put
> ```
> 
> I hope that's a good argument for putting it as close to the core as possible.

Definitely. and microlens really isnt a big package and would really clean up a lot of base XMonad functions.

slotThe:
On Wed, Feb 03 2021 23:21, AusCyberman wrote:
> Definitely. and microlens really isnt a big package and would really
> clean up a lot of base XMonad functions.

As I said, you don't even need microlens at all if you only use the
Lens' type, as that's easily defined in the module that the lenses are
supposed to be in.  You also have microlens-th as an added dependency
but don't use it; I'm not sure why.

N.b.: I think it's customary to just name these manual lenses `field1L`
or `field1T` (for traversables), not a full `field1Lens`
resp. `field1Traversable`

EDIT: All of the trivial lenses like `^.`, `.~`, `%~`, and `%%~` are also readily defined standalone

slotThe:
> As I said, you don't even need microlens at all if you only use the Lens' type, as that's easily defined in the module that the lenses are supposed to be in. You also have microlens-th as an added dependency but don't use it; I'm not sure why. 

As a proof-of-concept, I have create an `XMonad.Lens` module without extra dependencies [here](https://gist.github.com/slotThe/2a1747debcbcfa99543a71352ee145ec).  This is all modulo some `INLINE` pragmas, proper infix priorities, and a few optimizations with `Coercible` (plus perhaps some more type aliases), but these are easy enough to add and would probably only add noise at this point.

ghost:
> > As I said, you don't even need microlens at all if you only use the Lens' type, as that's easily defined in the module that the lenses are supposed to be in. You also have microlens-th as an added dependency but don't use it; I'm not sure why.
> 
> As a proof-of-concept, I have create an `XMonad.Lens` module without extra dependencies [here](https://gist.github.com/slotThe/2a1747debcbcfa99543a71352ee145ec). This is all modulo some `INLINE` pragmas, proper infix priorities, and a few optimizations with `Coercible` (plus perhaps some more type aliases), but these are easy enough to add and would probably only add noise at this point.

If you like, I've already written up a decent amount of the documentation [here](https://github.com/wygulmage/xmonad/blob/optics_1_setters/src/XMonad/Internal/Optics.hs). If it's useful feel free to use. (Sorry 

We may also want to get @ekmett's permission (even though the definitions look simple now, a lot of must have gone into creating them).

Finally, please include a `screensL` lens like [this](https://github.com/wygulmage/xmonad/blob/optics_1_setters/src/XMonad/StackSet.hs):
```Haskell
import Data.List.NonEmpty (NonEmpty ((:|)))
-- ...
screensL ::
    (Functor m)=>
    (NonEmpty (Screen i l a sid sd) -> m (NonEmpty (Screen i l a sid' sd'))) ->
    StackSet i l a sid sd -> m (StackSet i l a sid' sd')
{- ^
@screensL@ is a @Lens@ from a 'StackSet' to a 'NonEmpty' list of 'Screen's. The head of the list is the 'current' 'Screen' and the tail is the 'visible' 'Screen's.
Use @stackSet '&' screensL '.~' scrs'@ to set the 'Screens' in @stackSet@ to @scrs'@.
Use @stackSet '&' screensL . 'traverse' '%~' f@ to map a function @f@ over all the 'Screens' in @stackSet@.
Use @stackSet & screensL . traverse '%%~' f@ to traverse an 'Applicative' function over the 'Screen's in @stackSet@.
-}
screensL f stackSet =
    fmap
    (\ ~(cur :| vis) -> stackSet{ current = cur, visible = vis })
    (f (current stackSet :| visible stackSet))
```

There is a ton of code in xmonad that traverses a function over all visible screens, including the current one.

geekosaur:
Perhaps this should happen post-release; this seems like the wrong time to be contemplating rewriting things in terms of lenses. :)

Also, not clear to me how much "permission" for the above would want to be from van Laarhoven; @ekmett didn't invent lenses, he just made them more useful and popularized them.

ghost:
> Perhaps this should happen post-release; this seems like the wrong time to be contemplating rewriting things in terms of lenses. :)
> 
> Also, not clear to me how much "permission" for the above would want to be from van Laarhoven; @ekmett didn't invent lenses, he just made them more useful and popularized them.

@twanvl came up with the `Functor`-based implementation, but I don't know how many of the _operators_ used in the pre-`Profunctor` lens are his (including `#.`-based optimizations).

slotThe:
@wygulmage I'm definitely up for incorporating more lenses into this, as well as docs that are already there! That gist I posted isn't supposted to be "the start of rewriting xmonad entirely with lenses" or anything like that; just a small proof-of-concept, as I said :)

For what it's worth I wrote that module from memory; but of course that memory has been irrevocably contaminated by VL and profunctor-based (though that's not an option in this case of course) lenses!  
I do think that some finer details of the performance work are substantial enough that we should give credit to the respective authors, but I'd say we'll cross that bridge when we get to it—this is still quite a way away from making it into the core, after all.

> Perhaps this should happen post-release; this seems like the wrong time to be contemplating rewriting things in terms of lenses. :)

This is a 100% yes from me; I wouldn't even dream of rewriting the core that much so close to a release.

ghost:
Yep, all I meant is that I'd want the optics to live as close as possible to the core of xmonad because the core code _could_ get a lot of benefit from them, not that now is the time to refactor everything to use optics.

elkowar:
Definitely a +1 from me for adding some sort of lenses. Working with the StackSet can get _really_ tedious from time to time, this'd definitely help a lot

ekmett:
> We may also want to get @ekmett's permission (even though the definitions look simple now, a lot of must have gone into creating them).

You hereby have carte blanche from me to use the ideas and implementations from lens as you see fit under whatever open source license you need.

auscyber:
I feel like, there shouldn't really be labels, because, say for an xmonad-contrib module that needs other parts of microlens, the types might conflict. microlens doesnt have any external dependencies other than transformers. which xmonad needs anyway

slotThe:
On Mon, Feb 15 2021 16:29, AusCyberman wrote:
> I feel like, there shouldn't really be labels, because, say for an
> xmonad-contrib module that needs other parts of microlens, the types
> might conflict. microlens doesnt have any external dependencies other
> than transformers. which xmonad needs anyway

But microlens itself also needs to be built.  And the types do in fact
_not_ conflict with each other, because they are just transparent type
aliases.  Should you import lens or microlens in a personal project you
need only hide the combinators from one of the two imports (the xmonad
lens or the library lens implementation)—nothing will conflict at all.

I do think it's best to ice this for now and then, when we have a new
xmonad and xmonad-contrib release, create a self-contained lens
implementation and then decide whether we even want that at all.

elkowar:
I might not be in the best position to judge this, but still:
I'd argue that adding microlens as a dependency to XMonad core isn't really a dealbreaker. I doubt that many people - if any - use XMonad without XMonad-contrib, so adding a single dependency doesn't really seem fatal to me. The _practical_ impact on build-times seems negligible.
Then again, I might of course be missing something or just be wrong in my assumption here, please do correct me if I'm wrong.

Doing a self-contained lens implementation seems like a fine alternative, although just kind of unnecessary to me - the fact that people may already be familiar with the exact API of microlens  should definitely be considered. Although we could of course just replicate that closely. 

Waiting until _after_ the next release I definitely do agree with though, it would be a bad idea to make a change like this right now.

auscyber:
Microlens TH, could probably also be added to xmonad-contrib for Lenses for custom datatypes

ghost:
What are your thoughts on using polymorphic type classes for `Workspace` lenses, like this:

```haskell
  class HasLayout a a' l l' | a -> l, a' -> l', a l' -> a', a' l -> a where
     layoutL :: (Functor m)=> (a -> m a') -> l -> m l'
```
?

It would simplify ``currentL . workspaceL . stackL`` to ``currentL . stackL``, etc. Which is actually a decent gain in keeping the API small and avoiding cognitive noise, especially if the classes are not exported.

slotThe:
This looks like a decent gain in clarity, so we should definitely consider it I think.  The fundeps make me a bit sceptical though

ghost:
> This looks like a decent gain in clarity, so we should definitely consider it I think. The fundeps make me a bit sceptical though

The fundeps are necessary for type inference, and give a nice description of what the lens can do (although it's a bit of a pun): If you have an `a` you can use `layoutL` to get an `l`; if you have an `a'` you can use `layoutL` to get an `l'`; if you have an `a` (which contains an `l`) and an `l'`, you can use `layoutL` to make an `a'`, etc.

If the xmonad types weren't so polymorphic, these would be way simpler (e.g. `class HasLayout a where layoutL :: (Functor m)=> (Layout -> m Layout) -> a -> m a` -- no mult-param type class, no fundeps).

But you need the fully polymorphic version to be able to wrap or unwrap an instance of `LayoutClass` with the `Layout` constructor.

solomon-b:
I'm pretty late to this conversation but for whatever its worth I would like to throw out a few observations from experience.

1. Lenses have a tendency to multiply in a codebase.
2. It is easy catch yourself writing rather unreadable code that could have been more easily understood without lenses.

That said I love using lenses and find them to be a really beautiful structure that can be extremely effective when used appropriately. 

The `xmonad` and `xmonad-contrib` codebases are really old and have a lot of idiosyncrasies. If it were up to me (which it is not!) then I would suggest we first work to clean up the code base and go after low hanging fruit in terms of structure and readability and then decide if it makes sense to bring in lenses.

geekosaur:
For what it's worth, we've already started to do that kind of cleanup but only a little bit of it will make it into 0.17. Patches gratefully accepted, though!

geekosaur:
Also we have to be careful about such cleanups: we've already had to revert one because it broke other things that depended on the apparently idiosyncratic and wasteful behavior.

solomon-b:
> For what it's worth, we've already started to do that kind of cleanup but only a little bit of it will make it into 0.17. Patches gratefully accepted, though!

I'm working on a redesign on Prompt completions at the moment! 

I agree you have to be really careful refactoring old code as is the case here.

PR #274, opened 2021-02-16
Pass 0 to setClientMessageEvent where appropriate

liskin:
### Description

Stop pretending it has anything to do with time, it's zero, and it's not
being read anywhere, we're passing the parameter only because
setClientMessageEvent needs one. We should switch to
setClientMessageEvent' and pass no data, but that would move the X11
dependency lower bound.

### Checklist

### DIFF
--- a/src/XMonad/Main.hs
+++ b/src/XMonad/Main.hs
@@ -133,7 +133,7 @@ sendRestart = do
     xmonad_restart <- internAtom dpy "XMONAD_RESTART" False
     allocaXEvent $ \e -> do
         setEventType e clientMessage
-        setClientMessageEvent e rw xmonad_restart 32 0 currentTime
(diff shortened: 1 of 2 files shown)


slotThe:
> We should switch to setClientMessageEvent' and pass no data, but that would move the X11 dependency lower bound.

I wonder if this is so relevant?  Less lying is always good, and move a dependency bound around seems like a small price to pay.

liskin:
Well the idea was that I have no idea when X11 updates in Stackage LTS and while people are somewhat used to building xmonad/xmonad-contrib from git, X11 is just a dependency that they might prefer to take from LTS or the distro or something. But then I think I later decided to just release X11 asap (yeah I know …) so that by the time we get to xmonad release, this isn't an issue and both this and #273 can be merged without worry.

Issue #199, opened 2019-09-08
when using cabal-install-3 to build xmonad, must add `--lib`


jwaldmann:
This is more a matter of documentation. It cost me some time to debug, so I am noting it here:

If you built xmonad  using `cabal-install-3` (with default config file),
then  `ghci .xmonad/xmonad.hs`, or `xmonad --recompile`, produces errors like
```
.xmonad/xmonad.hs:1:1: error:
    Could not load module ‘XMonad’
    It is a member of the hidden package ‘xmonad-0.15’.
```

The solution (?) is to un-hide the packages with `cabal install --lib xmonad xmonad-contrib`. Then `ghc` picks them up without any `-package`  options.

It seems `cabal-install` v3 currently is under-documented (e.g., https://www.haskell.org/cabal/users-guide/# seems to refer to version 2.4)

geekosaur:
The right way to use cabal-install in v2+ mode is probably a build script, not `cabal install` as such; stick to `v1-install` otherwise, as this will be only the start of the problems (as soon as there are external dependencies things will become more complex, for example).

Memorytaco:
You could also manually add an entry to ghc environment file( which is usually located at `~/.ghc/{x86_64-linux-8.8.1}/environments/default`. Try to find your specific location in your home directory. 

This environment file acts like a reference file for installed package which is located in `.cabal/store`.

jwaldmann:
Yes, `cabal install --lib` writes (adds) to the global env file. 

But using the global env is not recommended. We can ` cabal install --lib --package-env=xmonad` but then for `xmonad --recompile` to work: how does it know about this environment file?

Similar: https://github.com/tidalcycles/Tidal/issues/572 . In fact, using the global env for both xmonad and tidal is not working for me (because of conflicting versions for dependencies)

hvr:
@jwaldmann 

> But using the global env is not recommended. We can ` cabal install --lib --package-env=xmonad` but then for `xmonad --recompile` to work: how does it know about this environment file?

That's actually not that hard; `xmonad` has support for a `~/.xmonad/build` script which if present in the filesystem (& executable) will be used instead of the hard-wired `ghc` invocation (see https://github.com/xmonad/xmonad/blob/c54e7088f003228e31096e68eabc75d50d227c1c/src/XMonad/Core.hs#L681-L692)

I have a `.xmonad/build`-script that looks like

```bash
#!/bin/sh -e

if [ "$#" != 1 ]; then
[...]

PR #278, opened 2021-03-18
Add New Tutorial

slotThe:
### Description

So, as discussed on IRC, here is the first draft of the new tutorial.  It's open right now how we want the format to be (discussed were things like putting it on the new website, on a wiki, or even on a "read the docs"-type site), but this is more for the actual content of the tutorial than any layout subtleties.  

All comments, suggestions, and general feedback is very much welcome!  In particular the split between the (at this point unreleased) new versions of xmonad and xmonad-contrib got quite large at some point—much larger than I originally anticipated.  I tried my best to test everything on both versions, but please do tell if anything does not work.

The document itself has, among other things, a few questions I'd like other people to answer.  Replicated here for clarity; bold questions are doubly important.

  * **Is this comprehensible for people who don't know Haskell?**

  * **Is this comprehensible for people who are actually new to xmonad?**

  * **Is anything missing that beginners really ought to know about?**

  * Lots of code is very similar here, only sporting small adjustments.
    It might be more natural working with `GNU diff` files here, instead
    of copy-pasting slightly changed code over and over again.  This may
    or may not confuse beginners, needs feedback.

  * I wonder if the xmobar config should use font-awesome or something
    similar or if that's overdoing it.  People really like eye-candy.
    On that note, the xmobar configuration might be a little too much
    (the xmonad `PP` is basically a copy-paste of my personal settings),
    but perhaps this will give people the start they need and then they
    can focus on more important things than how their bar looks :)

  * Is `xscreensaver` still relevant?

Note to any people who may be new to GitHub, but still want to offer their opinion: If you want to view this rendered on GitHub directly you can go to "Files changed", click on the little `...` menu, and then go to "View file".  You can also send all feedback to me directly on IRC or via mail to `soliditsallgood@mailbox.org` if you don't want to create a GitHub account just for this.

### Checklist

  - [ ] I've confirmed these changes don't belong in xmonad-contrib instead
    Arguably, the tutorial should be here instead of in the contrib repo because core gets a lot more traffic and hence more people could spot it.

  ~~- [ ] I tested my changes with [xmonad-testing](https://github.com/xmonad/xmonad-testing)~~

    TODO

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,0 +1,180 @@
+# Install XMonad
+
+## Stack
+
+### Preparation
+
+We'll use the [XDG] directory specifications here, meaning our
+configuration will reside within `$XDG_CONFIG_HOME`, which is
+`~/.config` on most systems.  Let's create this directory and move to
+it:
+
+``` shell
+  $ mkdir -p ~/.config/xmonad && cd ~/.config/xmonad
+```
+
+If you already have an `xmonad.hs` configuration, you can copy it over
+now.  If not, you can use the defaults: create a file called `xmonad.hs`
+with the following content:
+
+``` haskell
+  import XMonad
+
+  main :: IO ()
+  main = xmonad def
+```
+
+### Install Stack
+
+The easiest way to get [stack] is probably via your system's package
+manager.  For example, on Debian:
+
+``` shell
+  $ apt install haskell-stack
+```
+
+If your distribution does not package stack, you can also easily install
+it via the following command (this is the recommended way to install
+stack via its [documentation][stack]):
+
+``` shell
+  $ curl -sSL https://get.haskellstack.org/ | sh
+```
+
+Yet another way would be via [ghcup]; this is similar to installers like
+`rustup`, in case you prefer that.
+
+### Create a New Project
+
+Let's create a stack project.  Since we're already in the correct
+directory (`~/.config/xmonad`), we can start by cloning the `xmonad` and
+the `xmonad-contrib` repositories:
+
+``` shell
+  $ git clone https://github.com/xmonad/xmonad
+  $ git clone https://github.com/xmonad/xmonad-contrib
+```
+
+This will give you the latest `$HEAD`; if you want you can also check
+out a tagged release, e.g.:
+
+``` shell
+  $ git clone --branch v0.16 https://github.com/xmonad/xmonad
+  $ git clone --branch v0.17 https://github.com/xmonad/xmonad-contrib
+```
+
+Starting a new stack project is as simple as running `stack init`.
+Stack should now inform you that it will use the relevant `stack` and
+`cabal` files from `xmonad` and `xmonad-contrib` to generate its
+`stack.yaml` file.  At the time of writing, this looks a little bit like
+this:
+
+```
+  $ stack init
+  Looking for .cabal or package.yaml files to use to init the project.
+  Using cabal packages:
+  - xmonad-contrib/
+  - xmonad/
+
+  Selecting the best among 19 snapshots...
+
+  * Matches https://raw.githubusercontent.com/commercialhaskell/stackage-snapshots/master/lts/17/9.yaml
+
+  Selected resolver: https://raw.githubusercontent.com/commercialhaskell/stackage-snapshots/master/lts/17/9.yaml
+  Initialising configuration using resolver: https://raw.githubusercontent.com/commercialhaskell/stackage-snapshots/master/lts/17/9.yaml
+  Total number of user packages considered: 2
+  Writing configuration to file: stack.yaml
+  All done.
+```
+
+If you look into your current directory now, you should see a freshly
+generated `stack.yaml` file:
+
+```
+  $ ls
+  xmonad  xmonad-contrib  stack.yaml  xmonad.hs
+```
+
+The meat of that file (comments start with `#`, we've omitted them here)
+will look a little bit like
+
+``` yaml
+  resolver:
+    url: https://raw.githubusercontent.com/commercialhaskell/stackage-snapshots/master/lts/17/9.yaml
+
+  packages:
+  - xmonad
+  - xmonad-contrib
+```
+
+### Install Everything
+
+Installing things is as easy as typing `stack install`.  This will
+install the correct version of GHC, as well as build all of the required
+packages (`stack build`) and then copy the relevant executables
+(`xmonad`, in our case) to `~/.local/bin`.  Make sure to add that
+directory to your `$PATH`!
+
+If you're getting build failures while building the `X11` package it may
+be that you don't have the required C libraries installed.  Depending on
+your system, this may be `libX11-devel`, or `libxss`.
+
+### Tell XMonad How to Recompile Itself
+
+In order to tell xmonad to invoke `stack build` when we issue `xmonad
+--recompile` (bound to `M-q` by default), we need to create a so-called
+`build` file.  This is quite literally just a shell script called
+`build` in your xmonad directory (which is `~/.config/xmonad` for us)
+that tells xmonad how it should build its executable.
+
+A good starting point (this is essentially [what xmonad would do]
+without a build file, with the exception that we are invoking `stack
+ghc` instead of plain `ghc`) would be
+
+``` shell
+  #!/bin/sh
+
+  exec stack ghc --  \
+    --make xmonad.hs \
+    -i               \
+    -ilib            \
+    -fforce-recomp   \
+    -main-is main    \
+    -v0              \
+    -o "$1"
+```
+
+Don't forget to mark the file as `+x`: `chmod +x build`!
+
+And that's it!  Recompilation should work normally now, though you will
+potentially need to restart your computer, or at least the running X
+session, first.
+
+### Don't Recompile on Every Startup
+
+By default, xmonad always recompiles itself when a build script is used
+(because the build script could contain arbitrary code, so a simple
+check whether the `xmonad.hs` file changed is not enough).  If you find
+that too annoying, then you can use the `xmonad-ARCH` executable that
+`xmonad --recompile` generates instead of `xmonad` in your startup.  For
+example, instead of writing
+
+``` shell
+  exec xmonad
+```
+
+in your `~/.xinitrc`, you would write
+
+``` shell
+  exec $HOME/.local/share/xmonad/xmonad-x86_64-linux
+```
+
+The `~/.local/share` prefix is the `$XDG_DATA_DIR` directory.  Note that
+if your xmonad configuration resides within `~/.xmonad`, then the
+executable will also be within that directory and not in
+`$XDG_DATA_DIR`.
+
+[XDG]: https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
+[stack]: https://docs.haskellstack.org/en/stable/README/
+[ghcup]: https://www.haskell.org/ghcup/
+[what xmonad would do]: https://github.com/xmonad/xmonad/blob/master/src/XMonad/Core.hs#L657-L665
--- a/tutorial.md
+++ b/tutorial.md
@@ -740,23 +740,33 @@ and then plug that into our main function:
 ```
 
 As before, we now change things by modifying that `def` record until we
-find something that we like.  There are _a lot_ of options for the [PP
-record]; I'd advise you to read through all of them now, so you don't
-get lost!
+find something that we like.  First, for some functionality that we need
+further down we need to import one more module:
+
+``` haskell
+  import XMonad.Util.Loggers
+```
+
+Now we are finally ready to make things pretty.  There are _a lot_ of
+options for the [PP record]; I'd advise you to read through all of them
+now, so you don't get lost!
 
 ``` haskell
   myXmobarPP :: PP
   myXmobarPP = def
       { ppSep             = magenta " • "
-      , ppTitle           = wrap (white    "[") (white    "]") . magenta . ppWindow
-      , ppTitleUnfocused  = wrap (lowWhite "[") (lowWhite "]") . blue    . ppWindow
       , ppTitleSanitize   = xmobarStrip
       , ppCurrent         = wrap " " "" . xmobarBorder "Top" "#8be9fd" 2
       , ppHidden          = white . wrap " " ""
       , ppHiddenNoWindows = lowWhite . wrap " " ""
       , ppUrgent          = red . wrap (yellow "!") (yellow "!")
+      , ppOrder           = \[ws, l, _, wins] -> [ws, l, wins]
+      , ppExtras          = [logTitles formatFocused formatUnfocused]
       }
     where
+      formatFocused   = wrap (white    "[") (white    "]") . magenta . ppWindow
+      formatUnfocused = wrap (lowWhite "[") (lowWhite "]") . blue    . ppWindow
+
       -- | Windows should have *some* title, which should not not exceed a
       -- sane length.
       ppWindow :: String -> String
@@ -771,12 +781,12 @@ get lost!
       lowWhite = xmobarColor "#bbbbbb" ""
 ```
 
-_IF YOU ARE ON A VERSION `< 0.17`_: Both `ppTitleUnfocused` and
-  `xmobarBorder` are not available yet, so you will have to remove them.
-  As an alternative to `xmobarBorder`, a common way to "mark" the
-  currently focused workspace is by using brackets; you can try
-  something like `ppCurrent = wrap (blue "[") (blue "]")` and see if you
-  like it.
+_IF YOU ARE ON A VERSION `< 0.17`_: Both `logTitles` and `xmobarBorder`
+  are not available yet, so you will have to remove them.  As an
+  alternative to `xmobarBorder`, a common way to "mark" the currently
+  focused workspace is by using brackets; you can try something like
+  `ppCurrent = wrap (blue "[") (blue "]")` and see if you like it.  Also
+  read the bit about `ppOrder` further down!
 
 That's a lot!  But don't worry, take a deep breath and remind yourself
 of what you read above in the documentation of the [PP record].  Even if
@@ -787,8 +797,68 @@ focused window, `ppCurrent` formats the currently focused workspace,
 The rest is just deciding on some pretty colours and formatting things
 just how we like it.
 
-If this is too much for you, you can also really just start with the
-blank
+An important thing to talk about may be `ppOrder`.  Quoting from its
+documentation:
+
+  > By default, this function receives a list with three formatted
+  > strings, representing the workspaces, the layout, and the current
+  > window title, respectively. If you have specified any extra loggers
+  > in ppExtras, their output will also be appended to the list.
+
+So the first three argument of the list (`ws`, `l`, and `_` in our case,
+where the `_` just means we want to ignore that argument and not give it
+a name) are the workspaces to show, the currently active layout, and the
+title of the focused window.  The last element—`wins`—is what we gave to
+`ppExtras`; if you added more loggers to that then you would have to add
+more items to the list, like this:
+
+``` haskell
+  ppOrder = \[ws, l, _, wins, more, more2] -> [ws, l, wins, more, more2]
+```
+
+However, many people want to show _all_ window titles on the currently
+focused workspace instead.  For that, one can use `logTitles` from
+[XMonad.Util.Loggers] (remember that module we just imported?).
+However, `logTitles` logs _all_ titles.  Naturally, we don't want to
+show the focused window twice and so we suppress it here by ignoring the
+third argument of `ppOrder` and not returning it.  The functions
+`formatFocused` and `formatUnfocused` should be relatively self
+explanitory—they decide how to format the focused resp. unfocused
+windows.
+
+By the way, the `\ ... ->` syntax in there is Haskell's way to express a
+[lambda abstraction] (or anonymous function, as some languages call it).
+All of the arguments of the function come directly after the `\` and
+before the `->`; in our case, this is a list with exactly four elements
+in it.  Basically, it's a nice way to write a function inline and not
+having to define it inside e.g. a `where` clause.  The above could have
+also be written as
+
+``` haskell
+  myXmobarPP :: PP
+  myXmobarPP = def
+      { -- stuff here
+      , ppOrder = myOrder
+        -- more stuff here
+      }
+    where
+      myOrder [ws, l, _, wins] = [ws, l, wins]
+      -- more stuff here
+```
+
+If you're unsure of the number of elements that your `ppOrder` will
+take, you can also specify the list like this:
+
+``` haskell
+  ppOrder = \(ws : l : _ : wins : _) -> [ws, l, wins]
+```
+
+This says that it is a list of _at least_ four elements (`ws`, `l`, the
+unnamed argument, and `wins`), but that afterwards everything is
+possible.
+
+This config is really quite complicated.  If this is too much for you,
+you can also really just start with the blank
 
 ``` haskell
   myXmobarPP :: PP
@@ -948,6 +1018,7 @@ this:
   import XMonad.Hooks.StatusBar.PP
 
   import XMonad.Util.EZConfig
+  import XMonad.Util.Loggers
   import XMonad.Util.Ungrab
 
   import XMonad.Layout.Magnifier
@@ -991,15 +1062,18 @@ this:
   myXmobarPP :: PP
   myXmobarPP = def
       { ppSep             = magenta " • "
-      , ppTitle           = wrap (white    "[") (white    "]") . magenta . ppWindow
-      , ppTitleUnfocused  = wrap (lowWhite "[") (lowWhite "]") . blue    . ppWindow
       , ppTitleSanitize   = xmobarStrip
       , ppCurrent         = wrap " " "" . xmobarBorder "Top" "#8be9fd" 2
       , ppHidden          = white . wrap " " ""
       , ppHiddenNoWindows = lowWhite . wrap " " ""
       , ppUrgent          = red . wrap (yellow "!") (yellow "!")
+      , ppOrder           = \[ws, l, _, wins] -> [ws, l, wins]
+      , ppExtras          = [logTitles formatFocused formatUnfocused]
       }
     where
+      formatFocused   = wrap (white    "[") (white    "]") . magenta . ppWindow
+      formatUnfocused = wrap (lowWhite "[") (lowWhite "]") . blue    . ppWindow
+
       -- | Windows should have *some* title, which should not not exceed a
       -- sane length.
       ppWindow :: String -> String
@@ -1178,6 +1252,7 @@ either :)
 [INSTALL.md]: https://github.com/xmonad/xmonad/blob/master/INSTALL.md#stack
 [XMonad.Config]: https://github.com/xmonad/xmonad/blob/master/src/XMonad/Config.hs
 [XMonad.ManageHook]: https://hackage.haskell.org/package/xmonad/docs/XMonad-ManageHook.html
+[XMonad.Util.Loggers]: https://hackage.haskell.org/package/xmonad-contrib/docs/XMonad-Util-Loggers.html
 [XMonad.Util.EZConfig]: https://hackage.haskell.org/package/xmonad-contrib/docs/XMonad-Util-EZConfig.html
 [XMonad.Layout.Renamed]: https://hackage.haskell.org/package/xmonad-contrib/docs/XMonad-Layout-Renamed.html
 [XMonad.Layout.Magnifier]: https://hackage.haskell.org/package/xmonad-contrib/docs/XMonad-Layout-Magnifier.html
@@ -1200,4 +1275,5 @@ either :)
 [Haskell]: https://www.haskell.org/
 [trayer-srg]: https://github.com/sargon/trayer-srg
 [record update]: http://learnyouahaskell.com/making-our-own-types-and-typeclasses
+[lambda abstraction]: https://wiki.haskell.org/Lambda_abstraction
 [GNU Emacs conventions]: https://www.gnu.org/software/emacs/manual/html_node/elisp/Key-Sequences.html#Key-Sequences
--- a/tutorial.md
+++ b/tutorial.md
@@ -1172,17 +1172,15 @@ everything in one line:
 
 ## Get in Touch
 
-The `freenode/#xmonad` channel is very friendly and helpful.  It is
-possible that people will not immediately answer—we do have lives as
+The `irc.libera.chat/#xmonad` channel is very friendly and helpful.  It
+is possible that people will not immediately answer—we do have lives as
 well, after all :).  Eventually though, people will usually chime in if
 they have something helpful to say; sometimes this takes 10 minutes,
 other times it may well take 10 hours.  If you don't have an IRC client
 ready to go, the easiest way to join is via [webchat]—just jot down a
-username and `#xmonad` as the channel and you should be good to go!
-There is a [log] of the channel available, in case you do have to
-disconnect at some point, so don't worry about missing any messages.  We
-also have a [matrix room] that's linked to IRC, in case you prefer
-that.
+username and you should be good to go!  There is a [log] of the channel
+available, in case you do have to disconnect at some point, so don't
+worry about missing any messages.
 
 If you're not a fan of real-time interactions, you can also post to the
 [xmonad mailing list] or the [xmonad subreddit].
@@ -1230,11 +1228,10 @@ either :)
   - [TheMC47](https://github.com/TheMC47/dotfiles/tree/master/xmonad/.xmonad)
 
 
-[log]: https://ircbrowse.tomsmeding.com/browse/xmonad
+[log]: https://ircbrowse.tomsmeding.com/browse/lcxmonad
 [EWMH]: https://specifications.freedesktop.org/wm-spec/wm-spec-1.3.html
 [ICCCM]: https://tronche.com/gui/x/icccm/
-[webchat]: https://webchat.freenode.net/
-[matrix room]: https://matrix.to/#/#freenode_#xmonad:matrix.org
+[webchat]: https://kiwiirc.com/nextclient/irc.libera.chat/?#xmonad
 [about xmonad]: https://xmonad.org/about.html
 [shell variable]: https://www.shellscript.sh/variables1.html
 [xmonad-testing]: https://github.com/xmonad/xmonad-testing
--- a/tutorial.md
+++ b/tutorial.md
@@ -1,0 +1,1203 @@
+NOTE: `xmonad 0.16` and `xmonad-contrib 0.17` are obviously not released
+yet.  But don't worry, if you're reading this then you're already a beta
+tester :) Since this guide is supposed to work on older versions of
+xmonad as well this is no problem however; please report any
+incompatibilities you notice and state the version of xmonad and
+xmonad-contrib that you tried to use.  If you use the git versions of
+xmonad and xmonad-contrib, you should be able to follow everything just
+fine.
+
+# XMonad Configuration Tutorial
+
+We're going to take you, step-by-step, through the process of
+configuring xmonad, setting up [xmobar] as a status bar, configuring
+[trayer-srg] as a tray, and making it all play nicely together.
+
+We assume that you have read the [xmonad guided tour] already.  It is a
+bit dated at this point but, because xmonad is stable, the guide should
+still give you a good overview of the most basic functionality.
+
+Before we begin, here are two screenshots of the kind of desktop we will
+be creating together.  In particular, a useful layout we'll conjure into
+being—the three columns layout from [XMonad.Layout.ThreeColumns] with
+the ability to magnify stack windows via [XMonad.Layout.Magnifier]:
+
+<p>
+<img alt="blank desktop" src="https://user-images.githubusercontent.com/50166980/111586872-c0eb0e00-87c1-11eb-9129-c2781bbbf227.png" width="550">
+<img alt="blank desktop" src="https://user-images.githubusercontent.com/50166980/111586885-c47e9500-87c1-11eb-8e8e-83d4d33c4c8d.png" width="550">
+</p>
+
+So let's get started!
+
+## Preliminaries
+
+First you'll want to install xmonad.  You can either do this with your
+system's package manager or via `stack`/`cabal`.  The latter may be
+necessary if your distribution does not package `xmonad` and
+`xmonad-contrib` (or packages a version that's very old) or you want to
+use the git versions instead of a tagged release.  You can find
+instructions for how to do this in [INSTALL.md].
+
+One more word of warning: if you are on a distribution that installs
+every Haskell package dynamically linked—thus essentially breaking
+Haskell—(Arch Linux being a prominent example) then we would highly
+recommend installing via `stack` or `cabal` as instructed in
+[INSTALL.md].  If you still want to install xmonad via your system's
+package manager, you will need to `xmonad --recompile` _every time_ a
+Haskell dependency is updated—else xmonad may fail to start when you
+want to log in!
+
+We're going to assume xmonad version `0.16` and xmonad-contrib version
+`0.17` here, though most of these steps should work with older versions
+as well.  When we get to the relevant parts, will point you to
+alternatives that work with at least xmonad version `0.15` and
+xmonad-contrib version `0.16`.  This will usually be accompanied by a
+big "_IF YOU ARE ON A VERSION `< 0.17`_", so don't worry about missing
+it!
+
+Throughout the tutorial we will use, for keybindings, a syntax very akin
+to the [GNU Emacs conventions] for the same thing—so `C-x` means "hold
+down the control key and then press the `x` key".  One exception is that
+in our case `M` will not necessarily mean Alt (also called `Meta`), but
+"your modifier key"; this is Alt by default, although many people map it
+to Super instead (I will show you how to do this below).
+
+This guide should work for any GNU/Linux distribution and even for BSD
+folks.  Because debian-based distributions are still rather popular, we
+will give you the `apt` commands when it comes to installing software.
+If you use another distribution, just substitute the appropriate
+commands for your system.
+
+To install xmonad, as well as some utilities, via `apt` you can just run
+
+``` shell
+  apt-get install xmonad libghc-xmonad-contrib-dev libghc-xmonad-dev suckless-tools
+```
+
+This installs xmonad itself, everything you need to configure it, and
+`suckless-tools`, which provides the application launcher `dmenu`.  This
+program is used as the default application launcher on `M-p`.
+
+If you are installing xmonad with `stack` or `cabal`, remember to _not_
+install `xmonad` and its libraries here!
+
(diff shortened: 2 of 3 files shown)


tutorial.md
```diff
@@ -0,0 +1,1022 @@
...
+`stack` or `cabal` is used for installing xmonad (again see
+[xmonad-testing] for more information), even for BSD folks.  Because
+debian-based distributions are still rather popular, I will give you the
+`apt` commands when it comes to installing software.  If you use another
+distribution, just substitute the appropriate commands for your system.
+
+To install xmonad, as well as some utilities, via `apt` you can just run
+
+``` shell
+  apt-get install xmonad libghc-xmonad-contrib-dev libghc-xmonad-dev suckless-tools
```
geekosaur, line 74: This is still correct, for what it's worth. It'll get old versions of things, especially given we've already missed the cutoff for the next Debian release. (Not that they seem to care; they've ignored a couple of attempts to contact them already over a packaging question.)

tutorial.md
```diff
@@ -0,0 +1,1022 @@
...
+
+This installs xmonad itself, everything you need to configure it, and
+`suckless-tools`, which provides the application launcher `dmenu`.  This
+program is used as the default application launcher on `M-p`.
+
+For the remainder of this document, I will assume that you are running a
+live xmonad session in some capacity.  If you have set up your
+`~/.xinitrc` as directed in the xmonad guided tour, you should be good
+to go!  If not, just smack an `exec xmonad` at the bottom of that file.
+
```
geekosaur, line 88: Is it worth pointing to the wiki for various desktop environment setups here? https://wiki.haskell.org/Xmonad/Basic_Desktop_Environment_Integration in particular because many distributions have switched to this in place of `~/.xsession` files.

tutorial.md
```diff
@@ -0,0 +1,1022 @@
...
+    xmonad-contrib that you tried to use.  If you use the git versions
+    of xmonad and xmonad-contrib, you should be able to follow
+    everything just fine.
+
+TODO:
+
+  * I just took the apt commands from the old guide and guessed a bit;
+    someone who actually uses a debian-derivative has to confirm them.
+
+  * Switch all of the links to contrib 0.17 when it's out.
```
TheMC47: We could use https://hackage.haskell.org/package/xmonad-contrib/docs/XMonad-Hooks-EwmhDesktops.html instead of pinging to a certain release, which makes hackage redirect to the latest release AFAIK

tutorial.md
```diff
@@ -0,0 +1,1022 @@
...
+``` haskell
+  main :: IO ()
+  main = xmonad $ def
+      { modMask = mod4Mask  -- Rebind Mod to the Super key
+      }
+```
+
+The two dashes are a comment to the end of the line.  Notice the curly
+braces; these stand for a [record update] in Haskell (records are
+sometimes called "structs" in C-like languages).  What it means is "take
```
geekosaur, line 174: I've confused people with this one in the past; there are a lot of people who know Python and/or Javascript but not C/C++ these days.

tutorial.md
```diff
@@ -0,0 +1,1022 @@
...
+xmonad, setting up a status bar with [xmobar], setting up a tray with
+[trayer-srg], and making it all play nicely together.
+
+I assume that you have read the [xmonad guided tour] already.  It is a
+bit dated at this point but, because xmonad is stable, the guide should
+still give you a good overview of the most basic functionality.
+
+Before we begin, here are two screenshots of the kind of desktop we will
+be creating together.  In particular, a useful layout we'll conjure into
+being—the three columns layout from [XMonad.Layout.ThreeColumns] with
```
TheMC47, line 22: None of the modules are getting linked correctly, and I'm not sure why

tutorial.md
```diff
@@ -0,0 +1,1022 @@
...
+
+Don't mind the dollar sign too much; it essentially serves to split
+apart the `xmonad` function and the `def { .. }` record update visually.
+It's superfluous in this example, but that will change soon enough so
+it's worth introducing it here as well.
+
+What if we wanted to add other keybindings?  Say I also want to bind
+`M-S-z` to lock my screen with the screensaver, `C-<Print>` to take a
+snapshot of one window, and `<Print>` to take a snapshot of the entire
+screen.  This can be achieved with the `additionalKeys` function.  My
```
geekosaur: Do we want to start with `additionalKeys` or with `additionalKeysP`? The latter is more popular, not least because it means you don't have to hardcode the `modMask` in the keybinding (you can use `M-`); there is no way to accomplish this with `additionalKeys` except by abstracting it into a separate binding.

tutorial.md
```diff
@@ -0,0 +1,1022 @@
...
+      [ ((mod4Mask .|. shiftMask, xK_z    ), spawn "xscreensaver-command -lock")
+      , ((controlMask           , xK_Print), spawn "sleep 0.2; scrot -s"       )
+      , ((0                     , xK_Print), spawn "scrot"                     )
+      ]
+```
+
+Did you notice the `0` in the `xK_Print` line? The first part of the
+`(0, xK_Print)` tuple states what modifier keys (ctrl, alt, etc.) have
+to be held down for a pattern to match.  For the `Print` key, we don't
+need anything to be held down, and the zero indicates that.  The `sleep`
```
geekosaur: `XMonad.Util.Ungrab`

tutorial.md
```diff
@@ -0,0 +1,1022 @@
...
+before it was fullscreened, it will also do so afterwards.  Some people
+(me included) really like this behaviour, as applications thinking
+they're in fullscreen mode tend to remove a lot of clutter (looking at
+you, Firefox).  However, because a lot of people explicitly do not want
+this effect (and some applications, like chromium, will misbehave and
+need some [Hacks] to make this work), I will show you how to add the
+relevant function to get "proper" fullscreen behaviour here.
+
+_IF YOU ARE ON A VERSION `< 0.17`_: The `ewmhFullscreen` function does
+  not exist yet.  Instead of it, you can try to add
```
geekosaur: I would write "in these versions" instead of "yet", unless you're also alluding to it not being in any released version yet. Then again, this tutorial won't be readily visible until said release.

tutorial.md
```diff
@@ -0,0 +1,1022 @@
...
+      }
+    `additionalKeysP`
+      [ ("M-S-z"    , spawn "xscreensaver-command -lock")
+      , ("C-<Print>", spawn "sleep 0.2; scrot -s"       )
+      , ("<Print>"  , spawn "scrot"                     )
+      ]
+```
+
+_IF YOU ARE ON A VERSION `< 0.17`_: The `xmobarProp` function does not
+  exist yet.  Instead of it, use `xmobar` and carefully read the part
```
geekosaur: Same here re "yet".

tutorial.md
```diff
@@ -0,0 +1,1022 @@
+NOTES:
+
+  * xmonad 0.16 and xmonad-contrib 0.17 are obviously not released yet.
+    If you're read this then you're an alpha tester, congratulations :)
+    Since this guide is supposed to work on older versions of xmonad
+    aswell this is no problem however; please report any
+    incomparabilities you noticed and state the version of xmonad and
```
jpotier:
```suggestion
    incompatibilities you noticed and state the version of xmonad and
```

tutorial.md
```diff
@@ -0,0 +1,1022 @@
...
+  Simply replace _all_ occurences of `XMonadLog` with `StdinReader`
+  below (don't forget the template!)
+
+## Configuring Xmobar
+
+Now, before this will work, we have to configure xmobar.  Here's a nice
+starting point.
+
+``` haskell
+  Config { overrideRedirect = False
```
geekosaur, line 567: It may be worth noting that this config file only _looks_ like Haskell code. I've seen people try to use e.g. expressions or bindings in it.

tutorial.md
```diff
@@ -0,0 +1,1022 @@
...
+
+Save the file to `~/.xmobarrc`.  Now you should be able to press `M-q`
+to reload xmonad; this should now display xmobar with your new
+configuration!
+
+It is also possible to completely configure xmobar in Haskell, just like
+xmonad.  If you want to know more about that, you can check out the
+[xmobar.hs] example in the official documentation.  For a more
+complicated example, you can also check out [jao's xmobar.hs] (he's the
+author of xmobar).
```
geekosaur: Maintainer. Andrea Rossato was the original author.

tutorial.md
```diff
@@ -0,0 +1,1022 @@
...
+
+First you'll want to install xmonad.  You can either do this with your
+systems package manager, or (a bit more advanced) via `stack`/`cabal`.
+You can find instructions for the latter in the [xmonad-testing]
+repository.  I'm going to assume xmonad version `0.16` and
+xmonad-contrib version `0.17` here, though most of these steps should
+work with older versions as well.  When we get to the relevant parts,
+will point you to alternatives that work with at least xmonad version
+`0.15` and xmonad-contrib version `0.16`.  This will usually be
+accompanied by a big fat "_IF YOU ARE ON A VERSION `< 0.17`_", so don't
```
jpotier:
```suggestion
accompanied by a big "_IF YOU ARE ON A VERSION `< 0.17`_", so don't
```

Language? I don't think it's something I'd be at ease reading in the xmonad manual.

tutorial.md
```diff
@@ -0,0 +1,1022 @@
...
+
+Of course substitute the wallpaper for one of your own.  If you like the
+one used above, you can find it [here](https://i.imgur.com/9MQHuZx.png).
+
+## Final Touches
+
+There may be some programs that you don't want xmonad to tile.  The
+classic example is Gimp; it pops up all sorts of new windows all the
+time, and they work best at defined sizes.  It makes sense for xmonad to
+ignore these kinds of windows.
```
geekosaur: Do we want to call this "ignoring", given that `doIgnore` is also a thing (and generally not what is intended)?

tutorial.md
```diff
@@ -0,0 +1,1022 @@
...
+note down a username and `#xmonad` as the channel and you should be good
+to go!  We also have a [matrix server] that's linked to IRC, in case you
+do not want to use IRC.
+
+If you're not a fan of real-time interactions, you can also post to the
+[xmonad mailing list] or the [xmonad subreddit].
+
+## Trouble?
+
+Check `~/.xsession-errors` or your distributions equivalent first.  If
```
geekosaur: "distribution's"

jpotier (commented):
partial review

tutorial.md
```diff
@@ -0,0 +1,1022 @@
...
+         , template = "%XMonadLog% }{ %alsa:default:Master% | %cpu% | %memory% * %swap% | %EGPF% | %date% "
+         }
+```
+
+First, we set the font to use for the bar, as well as the colors.  The
+position options are documented well on the [xmobar home page] or,
+alternatively, in the [quick-start.org] on GitHub.  The particular
+option of `TopW L 90` says to put the bar in the upper left of the
+screen, and make it consume 90% of the width of the screen (we need to
+leave a little bit of space for `trayer-srg`).
```
TheMC47: It might be worth mentioning the famous https://github.com/jaor/xmobar/issues/239#issuecomment-233206552 which allows trayer to be seamlessly embedded into xmobar

tutorial.md
```diff
@@ -0,0 +1,1022 @@
...
+customized and perhaps a little bit hard to replicate (some may rely on
+features only available in personal forks or git), may or may not be
+documented, and most aren't very pretty either :)
+
+  - [byorgey](https://github.com/byorgey/dotfiles)
+  - [geekosaur](https://github.com/geekosaur/xmonad.hs/tree/pyanfar)
+  - [liskin](https://github.com/liskin/dotfiles/tree/home/.xmonad)
+  - [psibi](https://github.com/psibi/dotfiles/tree/master/xmonad)
+  - [slotThe](https://gitlab.com/slotThe/dotfiles/-/tree/master/xmonad/.config/xmonad)
+  - [TheMC47](https://github.com/TheMC47/dotfiles/tree/master/.xmonad)
```
TheMC47: Thanks for motivating me to clean-up my dotfiles :smiley:

TheMC47 (commented):
Just dropping by to say this is awesome :tada:  It's an easy read, and I just have a couple of comments

tutorial.md
slotThe: Thanks!  

Yeah I'd expect most people to not actually run debian but a derivative thereof, so this will _probably_ be fine.  Really what I'd like to do is to tell people to only install cabal and/or stack with their distros package manager and use projects to install anyting Haskell related; sadly, though, I can't see that happening anytime soon...

tutorial.md
slotThe: I'm a bit wary linking to the wiki due to outdated information, but that particular site doesn't look too bad.  That also sounds like a good page to have when redoing the website (cc'ing @TheMC47 so more people have this on their radar)

geekosaur: I keep that particular entry up to date since I use it. Most recently a couple weeks ago when I installed the machine I'm using now.

tutorial.md
slotThe: Aha good point.  I doubt there are any analogues in these kinds of dynamically/even weakly typed languages, are there?

tutorial.md
slotThe: You are right—the only purpose that `additionalKeys` seems to serve in here is to show how much "nicer" `additionalKeysP` is.  Might as well cut out the middleman in this case

tutorial.md
geekosaur: In the case of JS, IIRC there's no difference between a[b] and a.b, so hashes act like structs/classes. I don't recall how python does it but it has a "classy" syntax.

tutorial.md
slotThe: I was thinking about this and I ultimately decided that it would perhaps be too confusing for a beginner.  But then again, people may be much more familiar with shell scripting than with Haskell so it may not be a big deal after all

tutorial.md
slotThe: Good point, ignoring does have a specific meaning and floating a window is definitly not it—especially when we can just say what we want to do with the window: make it float

quarkQuark:
*\*Here from the Reddit post. Not a regular contributor, but I enjoy proofreading things just a bit too much!\**
Well done with this! XMonad is definitely in need of a good written tutorial, and this fits the bill marvellously. I've learnt a few things from it myself! Here's my initial thoughts, though feel free to ignore them as the tutorial is already great:

- I don't think there's any problem at all with the code duplication. Readability over brevity.
- As for using font-awesome in xmobar, I think it's fine to leave it out of the main body. But it may be worth considering a short 'where to go from here if you want more eyecandy' ideas section at the end.
- If this is to be a community tutorial, not one on a personal website, then it may be better to avoid use of the pronouns 'I' and 'me'. I think this makes it seem more professional, and will make more sense particularly once others start contributing to it.
- On Arch Linux (which could be more popular than Debian-based distros amongst the target audience), the packages for XMonad and dmenu have different names. I don't think it's a problem that you've only included the commands for apt, but it may be worth noting that package names, as well as commands, vary.
I do think there's a very strong argument for just telling everyone to use stack. It would need to be accompanied by reasons, though. I was very resistant to the idea of installing anything at all without the pacman before Arch dynamic linking broke XMonad one too many times.
- I imagine most people will try XMonad while they have a desktop environment, and therefore a display manager, already installed. It may be worth having a paragraph on this, as it will probably be easier for a newcomer.
- For anyone inexperienced with the command line and shell variables, it may be helpful to clarify what you mean by `$XDG_CONFIG_HOME`.
- Thanks for including the term 'record change'. I'd been wondering what to Google to make my own structures like that!
- The term 'top-level' (used when explaining the `where` syntax) is something I remember having to Google when I was learning Haskell, so it might be worth explaining it. 'Local declarations' may also not have an obvious meaning to someone new to programming in general.
- This is a very minor point, but I think it may be more correct to replace 'figure' with 'reason' in "try to figure this out for yourself by just looking at the documentation." 'Figuring something out' usually refers to finding an unknown answer, rather than understanding a known answer.
- Some people won't understand what you mean by 'commute' when talking about combinators. It may not be necessary to change it though, as you also say that the order matters.
- Property-based logging sounds exciting! I've had problems with the older way of doing it, so I look forward to moving over.
- "You can of course monitor more than one if you like, see xmobar's weather monitor documentation for further details." COMMA SPLICING! You have committed a serious grammatical crime! You can replace the comma with basically any other punctuation mark of your choice. I'd probably use a colon.
- I'm curious as to your reason for using trayer over stalonetray. I was under the impression stalonetray was a lot more popular?
- It may be worth suggesting nitrogen as a more familiar alternative to feh?
- Do you mean `xprop` rather than `xprompt`?
- I didn't know about the distinction between `className` and `appName`. I think I'll find that quite helpful.
- It's good that you have a few references to the docs. The XMonad documentation is stellar, but all the types can be quite intimidating for the newcomer. Pointing out things like 'it tells us to import the module' will come a long way in helping people to see that there are plain English paragraphs as well!

slotThe:
@quarkQuark Thank you for the thorough response!  I pushed changes to
some of the easy fixes.  Some answers below are just reminders for me to
actually do things, so don't mind that.

>     * As for using font-awesome in xmobar, I think it's fine to leave
>       it out of the main body. But it may be worth considering a short
>       'where to go from here if you want more eyecandy' ideas section
>       at the end.

That's a good idea!  I'll try to think of something.

>     * If this is to be a community tutorial, not one on a personal
>       website, then it may be better to avoid use of the pronouns 'I'
>       and 'me'. I think this makes it seem more professional, and will
>       make more sense particularly once others start contributing to
>       it.

Ah yeah; I usually default to 'we' and 'us' and the like but I heard
that some people find that too impersonal.  I figured that a few
occurences of "the speaker (whoever it may be) talking to the person
reading the document directly" would perhaps brighten the mood a little
bit, but I do agree that this now reads a bit weird.

>     * On Arch Linux (which could be more popular than Debian-based
>       distros amongst the target audience), the packages for XMonad
>       and dmenu have different names. I don't think it's a problem
>       that you've only included the commands for apt, but it may be
>       worth noting that package names, as well as commands, vary.  I
>       do think there's a very strong argument for just telling
>       everyone to use stack. It would need to be accompanied by
>       reasons, though. I was very resistant to the idea of installing
>       anything at all without the pacman before Arch dynamic linking
>       broke XMonad one too many times.

It _is_ probably worth pointing out that Arch essentially broke Haskell
and thus especially these users should consider stack.  I still don't
quite know how big of a hurdle it is to create a stack project + a build
file.  I would say that there's enough documentation in `xmonad-testing`
but it's certainly harder than just using your distro's package
manager...

>     * I imagine most people will try XMonad while they have a desktop
>       environment, and therefore a display manager, already
>       installed. It may be worth having a paragraph on this, as it
>       will probably be easier for a newcomer.

I don't think I've have personally never used a display manager (at
least that I can recall), so help is very much appreciated here!

>     * "You can of course monitor more than one if you like, see
>       xmobar's weather monitor documentation for further details."
>       COMMA SPLICING! You have committed a serious grammatical crime!
>       You can replace the comma with basically any other punctuation
>       mark of your choice. I'd probably use a colon.

Haha, and I thought I got them all!  One always slips through.

>     * I'm curious as to your reason for using trayer over
>       stalonetray. I was under the impression stalonetray was a lot
>       more popular?

The old tutorial used trayer and since I don't personally use a system
tray, I just copied most of that part over.  I also think that both
@liskin and @TheMC47 use trayer instead of stalonetray so I was hoping
they could offer more of their insight if I didn't switch the program
under their feet :)

>     * It may be worth suggesting nitrogen as a more familiar
>       alternative to feh?

I don't have a problem with alternative suggestions per se, but I feel
we could then get into dangerous territory where we also have to talk
about `slock` or `xsecurelock` instead of `xscreensaver`, `import`
instead of `scrot`, and however many power managers there are out there.

>     * It's good that you have a few references to the docs. The XMonad
>       documentation is stellar, but all the types can be quite
>       intimidating for the newcomer. Pointing out things like 'it
>       tells us to import the module' will come a long way in helping
>       people to see that there are plain English paragraphs as well!

This is also an excellent point, I'll go through this again and see 
whether I can directly quote from the docs to perhaps alleviate this fear
even more

tutorial.md
sorki: Stumbled on https://wiki.haskell.org/Xmonad/Installing_xmonad which you might want to link here.

slotThe: That page looks a bit moldy, especially debian's `libghc6-xmonad-contrib-dev` :)  Though it would be a cool idea to have all of these kinds of installation options on the new website and then link to that page!

liskin:
> It _is_ probably worth pointing out that Arch essentially broke Haskell
> and thus especially these users should consider stack. I still don't
> quite know how big of a hurdle it is to create a stack project + a build
> file. I would say that there's enough documentation in `xmonad-testing`
> but it's certainly harder than just using your distro's package
> manager...

I think telling people to create an actual project with both stack.yaml and something.cabal is quite a hurdle, but telling them to `stack install xmonad xmonad-contrib` and then create a build file that invokes `stack exec ghc … xmonad.hs` is probably okay. And I think we should try to figure out a way for the `stack exec` part to be unnecessary, by e.g. capturing `GHC_ENVIRONMENT` and `GHC_PACKAGE_PATH` env variables during xmonad build and re-setting them when recompiling, so that we don't need to tell people to do anything special, and regardless of how they install xmonad, it will do the right thing The xmonad-testing approach is unnecessarily complicated for most people (including me!).

Also, there's also [this trick](https://github.com/xmonad/xmonad/issues/199#issuecomment-609433196) (the bottom part of that comment) to make the recompilation just work when installing via cabal.

> The old tutorial used trayer and since I don't personally use a system
> tray, I just copied most of that part over. I also think that both
> @liskin and @TheMC47 use trayer instead of stalonetray so I was hoping
> they could offer more of their insight if I didn't switch the program
> under their feet :)

Unfortunately I don't remember why I ended up using trayer instead of stalonetrayer, it's been quite a while.

tutorial.md
slotThe: That's fair enough; I probably got a bit too carried away there :)

TheMC47:
> The old tutorial used trayer and since I don't personally use a system
> tray, I just copied most of that part over. I also think that both
> @liskin and @TheMC47 use trayer instead of stalonetray so I was hoping
> they could offer more of their insight if I didn't switch the program
> under their feet :)
> 
> >

To be honest, I also don't have a reason why I'm using trayer. I actually started with stalonetray, then I wanted to embed the systray into xmonad, which made me find that bash script. I had exams so I didn't bother adapting it to stalonetray and just jumped ship

tutorial.md
```diff
@@ -0,0 +1,1039 @@
...
+You can find the names for special keys in the `EZConfig` documentation.
+
+I will cover setting up the screensaver later in this tutorial.
+
+The `unGrab` before running the `scrot -s` command tells xmonad to
+release its keyboard grab before `scrot -s` tries to grab the keyboard
+itself.  The little `*>` operator essentially just sequences two
+functions, i.e. `f *> g` says
+
+  > first to `f` and, discarding any result that `f` may have given me,
```
harwiltz: Typo, should say "first *do* `f`"

tutorial.md
```diff
@@ -0,0 +1,1039 @@
...
+
+This functionality is to be found in the [XMonad.Hooks.EwmhDesktops]
+module, so let's import it:
+
+``` haskell
+  import XMonad.Hooks.EwmhDesktops
+```
+
+We might also consider using the `ewmhFullscreen` combinator.  By
+default, a "fullscreened" application is still bound by it's window
```
harwiltz: Sorry to be annoying with grammar, but should be "bound by its window"

harwiltz (commented):
Thank you so much for your hard work! This is a massive improvement over the current xmonad tutorial. All of the concepts that confused me when I set up xmonad about a year ago where addressed here.

tutorial.md
slotThe: Nothing to worry about; thank you for pointing these out!

tutorial.md
slotThe: @TheMC47 I added a little note in [this](https://github.com/xmonad/xmonad/pull/278/commits/e6c2a670dda3504209a359ccf7cff9992fe01c37) fixup commit; do you think that's enough or should I go into a little bit more detail?

slotThe:
> I think telling people to create an actual project with both stack.yaml and something.cabal is quite a hurdle, but telling them to `stack install xmonad xmonad-contrib` and then create a build file that invokes `stack exec ghc … xmonad.hs` is probably okay. 
> 
> [...]
>
> Also, there's also [this trick](https://github.com/xmonad/xmonad/issues/199#issuecomment-609433196) (the bottom part of that comment) to make the recompilation just work when installing via cabal.

I would be a bit careful when recommending cabal, as afaik if people v3-install a library globally and then want to update that, they have to nuke their `ghc-pkg` data base; otherwise there's a world of trouble.  Does installing globally with stack have the same problem?  If not that might actually be a good recommendation on our part.

tutorial.md
TheMC47: I think it's fine :+1:

liskin:
> I would be a bit careful when recommending cabal, as afaik if people v3-install a library globally and then want to update that, they have to nuke their `ghc-pkg` data base; otherwise there's a world of trouble. Does installing globally with stack have the same problem? If not that might actually be a good recommendation on our part.

I wasn't suggesting to recommend any specific tool, just to give guidance how to use those tools (both cabal and stack) with xmonad in a way that works including recompilation.

I was hoping that `--package-env=$HOME/.xmonad/` is not “installing a library globally”, if it is, that's unfortunate, and we should look into it some more and find a better way.

Re stack: it does not have the same problem. Installing via stack globally, provided `.stack/global/stack.yaml` specifies just a resolver, as it should, means xmonad is built in a directory specific to the resolver and GHC version, and if any of that is later updated, it's all built again into another directory. And if the global `stack.yaml` does specify additional packages, like a newer version of xmonad for example, it's built elsewhere but still likely named such that changing the xmonad version doesn't result in any sort of conflict.

geekosaur:
The package environment is "global" insofar as it can be referenced from anywhere, but you have to use the same package environment to see anything in it.

slotThe:
To re-iterate, the following questions are still open and somewhat out of my comfort-zone, so I would appreciate any help I could get there:

  * People may be inclined to use a display manager, especially when
    coming from proper DEs.  I have no experience with those, so help
    would be appreciated as to what a section or paragraph mentioning
    this should actually contain.

  * ~~Arch broke Haskell by forcing everything to be dynamically linked,
    so it may be worth mentioning other installation methods more
    prominently.  Entire stack or cabal projects may be too much, and
    cabal v3 doesn't like installing libraries globally; needs testing
    if stack does.~~
  Recommending a stack project seems sensible enough, now included in a new `INSTALL.md`

I think that if these are solved the main bulk of the work is done and this would be in a mergeable state.

tutorial.md
arnamak: For whatever it’s worth, the equivalent JS term for this bit would be [Object destructuring](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#object_destructuring)

slotThe:
@liskin I added an `INSTALL.md` file for a (I think) decently easy guide to get a stack project going.  What do you think?

liskin:
@slotThe First look (I spent like 30 seconds on it because I'm just heading out) looks okay! I still think we should try to make xmonad clever enough not to require the build script, sooner or later, but most of the guide will stay and it looks great. Thanks a lot.

(I know I haven't answered the other questions here, I still intend to do so, it's just taking time to get to it, sorry.)

INSTALL.md
```diff
@@ -0,0 +1,171 @@
...
+
+And that's it!  Recompilation should work normally now, though you will
+potentially need to restart your computer, or at least the running X
+session, first.
+
+### Don't Recompile on Every Startup
+
+By default, xmonad always recompiles itself when a build script is used
+(because the build script could contain arbitrary code, so a simple
+check whether the `xmonad.hs` file changed is not enough).  If you find
```
oogeek, line 157: I am a bit confused by that. I think usually users will run `xmonad --recompile`, and then xmonad will recognize the build script and recompile. But this sentence may sound like every time `xmonad` starts, it will recompile. Maybe a bit clarification. Maybe something like
```
When running `xmonad --recompile`, xmonad will always ...
```

tutorial.md
```diff
@@ -0,0 +1,1140 @@
...
+We might also consider using the `ewmhFullscreen` combinator.  By
+default, a "fullscreened" application is still bound by its window
+dimensions; this means that if the window occupies half of the screen
+before it was fullscreened, it will also do so afterwards.  Some people
+(me included) really like this behaviour, as applications thinking
+they're in fullscreen mode tend to remove a lot of clutter (looking at
+you, Firefox).  However, because a lot of people explicitly do not want
+this effect (and some applications, like chromium, will misbehave and
+need some [Hacks] to make this work), we will also add the relevant
+function to get "proper" fullscreen behaviour here.
```
oogeek: The link of `Hacks` is broken. I think you are talking about the X.U.Hacks from the git version.

tutorial.md
```diff
@@ -0,0 +1,1140 @@
...
+One more word of warning: if you are on a distribution that installs
+every Haskell package dynamically linked—thus essentially breaking
+Haskell—(Arch Linux being a prominent example) then we would highly
+recommend installing via `stack` or `cabal` as instructed in
+[INSTALL.md].  If you still want to install xmonad via your system's
+package manager, you will need to `xmonad --recompile` _every time_ a
+Haskell dependency is updated.
+
+We're going to assume xmonad version `0.16` and xmonad-contrib version
+`0.17` here, though most of these steps should work with older versions
```
oogeek, line 51: Maybe also mention the git version here. I think xmonad-contrib 0.17 is not finalized. Also do we need to differentiate between 0.17 and git at this moment?

tutorial.md
```diff
@@ -0,0 +1,1140 @@
...
+  main :: IO ()
+  main = xmonad
+       . ewmhFullscreen
+       . ewmh
+     =<< statusBarProp "xmobar" def toggleStrutsKey myConfig
+    where
+      toggleStrutsKey :: XConfig Layout -> (KeyMask, KeySym)
+      toggleStrutsKey XConfig{ modMask = m } = (m, xK_b)
+```
+
```
oogeek: Maybe it is good to add an example with specifying the xmobar config path, such as
```
statusBarProp "xmobar -x 0 ~/.config/xmobar/xmobarrc0" def toggleStrutsKey myConfig
``` 
Can you provide a multiscreen example here?

INSTALL.md
slotThe: It sounds like I'm saying that because I'm saying exactly that! :)  When you use a build script we have no idea what it does and so we recompile _every time_ xmonad starts

tutorial.md
slotThe: I am talking about X.U.Hacks, but since this tutorial will come out with the new release, I've already switched it to the appropriate Hackage address (which currently doesn't exist) so we can't forget

tutorial.md
slotThe: The tutorial will be "released" in tandem with 0.17, so I think this still makes sense

tutorial.md
slotThe: This sounds like a good idea for the `Further Customizations` section, thanks!  I'd like to wait until monad/xmonad-contrib/pull/463 is finalized though, so we can showcase that functionality straight away.

TheMC47:
> * People may be inclined to use a display manager, especially when
>   coming from proper DEs.  I have no experience with those, so help
>   would be appreciated as to what a section or paragraph mentioning
>   this should actually contain.
> 

I'm using sddm, but I configured it some time ago and forgot what I did
exactly... I'll need to test this on a VM (which might take some time, I'm
still all over the place). But I think it's not really necessary, so maybe
we could merge this like this and it later?

slotThe:
> I'm using sddm, but I configured it some time ago and forgot what I did
> exactly... I'll need to test this on a VM (which might take some time, I'm
> still all over the place). 

No worries, take your time :)

> But I think it's not really necessary, so maybe
> we could merge this like this and it later?

Fine by me—as long as we remember to do what we wanted to do before releasing 0.16/0.17!  ~~I would perhaps like one more read through the whole thing to sort out some inconsistencies but then~~ we can squash all those fixups and merge. 

If you're so inclined, it would be cool if you found the time to perhaps look over [this commit](https://github.com/xmonad/xmonad/pull/278/commits/7d43128753f3a9a8128346034997e4c0b456058d), which tried to make the tutorial compliant with the new `X.H.StatusBar` stuff

oogeek:
> Fine by me—as long as we remember to do what we wanted to do before releasing 0.16/0.17! I would perhaps like one more read through the whole thing to sort out some inconsistencies but then we can squash all those fixups and merge.

Personally I use `startx`.
I am willing to help test out the display managers. 
So the `Arch` Install of `XMonad` already provides a `desktop file` put under `/usr/share/xsessions`, I have tested it and it works well. But that is for users who install XMonad from official repos. However, I think for others, with slight modifications, it will also work.

The unlucky thing is, both `sddm` and `lightdm` are not working on my Arch setup, so I only tested `gdm` from `Gnome` and it works well.

Here is the desktop file of interest:
```
xmonad.desktop
```


```
[Desktop Entry]
Version=1.0
Type=Application
Name=Xmonad
Comment=Lightweight X11 tiled window manager written in Haskell
Exec=xmonad
Icon=xmonad
Terminal=false
StartupNotify=false
Categories=Application;
```

Hope that it helps anyway.

TheMC47:
I can confirm I have the same desktop file with `sddm` (well, almost the same since I'm launching KDE too)

liskin:
Back when I used a desktop manager (lightdm), I launched xmonad via .xsession, and thus if I ever needed a second X session, startx just worked. But I've never used gdm, so it's entirely possible that a desktop file is a better experience with that.

tutorial.md
```diff
@@ -0,0 +1,1194 @@
...
+via the property (e.g. have `"<fn=1>\xf120</fn>"` as one of your
+workspace names).
+
+As an example how this would look like in a real configuration, you can
+look at [Liskin's], [slotThe's], or [TheMC47's] xmobar configuration.
+Do note that the last two are Haskell-based and thus may be a little
+hard to understand for newcomers.
+
+[Liskin's]: https://github.com/liskin/dotfiles/blob/home/.xmobarrc
+[TheMC47's]: https://github.com/TheMC47/dotfiles/blob/32ca65cacfc395873ced133102125263a276db50/.xmonad/mo.hs
```
TheMC47:
```suggestion
[TheMC47's]: https://github.com/TheMC47/dotfiles/tree/master/xmobar/xmobarrc
```

tutorial.md
```diff
@@ -0,0 +1,1194 @@
...
+hard to replicate (some may rely on features only available in personal
+forks or git), may or may not be documented, and most aren't very pretty
+either :)
+
+  - [byorgey](https://github.com/byorgey/dotfiles)
+  - [geekosaur](https://github.com/geekosaur/xmonad.hs/tree/pyanfar)
+  - [liskin](https://github.com/liskin/dotfiles/tree/home/.xmonad)
+  - [psibi](https://github.com/psibi/dotfiles/tree/master/xmonad)
+  - [slotThe](https://gitlab.com/slotThe/dotfiles/-/tree/master/xmonad/.config/xmonad)
+  - [TheMC47](https://github.com/TheMC47/dotfiles/tree/master/.xmonad)
```
TheMC47:
```suggestion
  - [TheMC47](https://github.com/TheMC47/dotfiles/tree/master/xmonad/.xmonad)
```

TheMC47 (commented):
I couldn't push to your branch; I updated my dotfiles and here are the new links

tutorial.md
```diff
@@ -645,15 +689,48 @@ Let's copy the implementation over into our main function:
   main = xmonad
        . ewmhFullscreen
        . ewmh
-     =<< statusBarProp "xmobar" def toggleStrutsKey myConfig
+       $ withEasySB (statusBarProp "xmobar" (pure def))
+                    defToggleStrutsKey
+                    myConfig
```
TheMC47: This is fine, though I would prefer having the all the status bar stuff as one modifier. This should give the users the impression that it's just another combinator that you could add and remove, so something like this:
```haskell
  main = xmonad
       . ewmhFullscreen
       . ewmh
       . withEasySB (statusBarProp "xmobar" (pure def)) defToggleStrutsKey
       $ myConfig
```
I think this puts more emphasis on the composability of the different modifiers, and just looks a bit nicer. But this is just a very minor nitpick

TheMC47 (commented):
Awesome work! Just a really minor nitpick, and it's very subjective

tutorial.md
slotThe: Oh that's brilliant, thanks!

tutorial.md
```diff
@@ -0,0 +1,1196 @@
...
+
+  import XMonad.Hooks.EwmhDesktops
+
+
+  main :: IO ()
+  main = xmonad
+       . ewmhFullscreen
+       . ewmh
+       . withEasySB (statusBarProp "xmobar" (pure myXmobarPP)) defToggleStrutsKey
+       $ myConfig
```
oogeek, line 1035: Also a nitpick, I think you can mention in comment that users can use "xmobar ~/path/to/config" so that it is clearer. Some users might just use "xmobar" and then ask question about why it does not work.

oogeek: Also, if they don't go and read the source code, they may say, oh, only `xmobar` should be here. But actually it can be a command line.

slotThe: > Also a nitpick, I think you can mention in comment that users can use "xmobar ~/path/to/config" so that it is clearer. Some users might just use "xmobar" and then ask question about why it does not work.

xmobar reads from `~/.xmobarrc` by default, and this is the path that we tell people to install the config in.  We could put it down to user error if they put it somewhere else, but perhaps you're right and we should mention this.

> Also, if they don't go and read the source code, they may say, oh, only `xmobar` should be here. But actually it can be a command line.

I personally don't know of another bar that fascilitates property-based logging by default, so that assumption may or may not be accurate :)  I think the fact that it's a string and this is configurable should be more-or-less obvious; otherwise it would surely be hardcoded!

oogeek: Yeah, I mean `xmobar` used with `command line options`. I agree with what you said. It would be OK to mention a bit of it. So do you intend to include a `user error` part in this PR, or in later PR?

slotThe:
On Sat, May 01 2021 08:37, oogeek wrote:
> Yeah, I mean `xmobar` used with `command line options`. I agree with
> what you said. So it would be OK to mention a bit of it. So do you
> intend to include a `user error` part in this PR, or in later PR?

I'll push a change later.  Thanks!

tutorial.md
slotThe: I added a little something

oogeek: Thanks, it looks nice.

oogeek:
@slotThe  So do you think it would be good to add a paragraph about starting xmonad:
1. use `startx` 
2. use display manager

I just post the .desktop file in the comment, it comes from Arch Package of xmonad, and I think it would be OK. :)

I mean, you have already mentioned a bit of `startx` in the `##Trouble?` part. So why not add a bit more?

slotThe:
>     1. use `startx`
>
> [...]
>
> I mean, you have already mentioned a bit of `startx` in the `##Trouble?` part. So why not add a bit more?

Because we tell people to put `exec xmonad` into their `~/.xinitrc`, xmonad should start up correctly on `startx`; what more is there to add?

>     2. use display manager
> 
> I just post the .desktop file in the comment, it comes from Arch Package of xmonad, and I think it would be OK. :)

I'm all for adding this, but like I said I have never used a display manager (and, on top of that, don't even know what a `.desktop` file is :)), so I doubt I could do a good job with explaining that part.  If someone else wants to write a few lines then do feel free to

tutorial.md
```diff
@@ -0,0 +1,1212 @@
...
+      ratio   = 1/2    -- Default proportion of screen occupied by master pane
+      delta   = 3/100  -- Percent of screen to increment by when resizing panes
+```
+
+The so-called `where`-clause above simply consists of local declarations
+that might clutter things up where they all declared at the top-level
+like this
+
+``` haskell
+  myLayout = Tall 1 (1/2) (3/100) ||| Mirror (Tall 1 (1/2) (3/100)) ||| Full
```
romanZupancic: A minor nitpick: this code snippet sees the `ratio` and `delta` parameters switched (delta is 1/2 and ratio is 3/100) for both Tall configurations.

This should really read:
```haskell
myLayout = Tall 1 (3/100) (1/2) ||| Mirror (Tall 1 (3/100) (1/2)) ||| Full
```

slotThe: Well spotted, thanks!

slotThe:
We also may need to wait for how /xmonad/xmonad-contrib/pull/535 goes, because `ppTitleUnfocused` is currently used in the xmobar configuration.

slotThe:
Okay I updated everything to use `logTitles` instead of `ppTitleUnfocused` in a separate commit, for ease of review.  I also converted this from its draft state to a "normal" pr so we can slowly think about merging it.

liskin:
Lest I forget, a couple issues that surfaced today:

* the tutorial isn't linked from README (and the README needs a facelift in general)
* possibly also consider uppercasing the filename to make it easier to find alongside README and INSTALL and LICENSE, this should be done early or not at all, before anyone links to the tutorial
* `overrideRedirect = False` disables `lowerOnStart`, which then breaks toggling of struts (repro: open a couple windows, restart xmobar, now xmobar is stacked on top of these and toggling avoidstruts results in windows being covered by xmobar) — do you remember why it's there? is it because https://github.com/xmonad/xmonad-contrib/pull/492 wasn't merged when the tutorial was written?

slotThe:
These are good points, thanks!  The first two should be taken care of by 40cd2081da08047f3b265419ab1d36430c3d083c.  

The third one is a bit of a head-scratcher.  I vaguely remember the flag helping with some problem I had regarding restarting xmonad and then "old" windows not being able to cover xmobar at all.  I have never had any problems with toggling struts with `overrideRedirect` being disabled, however.

liskin:
> I vaguely remember the flag helping with some problem I had regarding restarting xmonad and then "old" windows not being able to cover xmobar at all.

That flag is indeed related to this problem, but it's precisely the `overrideRedirect = False` setting that **causes** this problem, not prevents it. That is, if one uses `.xmobarrc`, because `overrideRedirect = False` implies `lowerOnStart = False`, and lowering on start is what's necessary for old windows to be able to cover xmobar. In a compiled configuration, `overrideRedirect` and `lowerOnStart` are completely separate, and that's why you personally haven't experienced any trouble.

What `overrideRedirect` affects in both parsed and compiled configurations is how ManageDocks detects it. When it's True, xmonad core knows not to manage the window and ManageDocks needs to rely on raw X events to detect the window and subsequent changes to its properties when the screen layout changes (and that was buggy prior to https://github.com/xmonad/xmonad-contrib/pull/492). When `overrideRedirect` is False, xmonad core treats it as a normal window and a manage hook is needed to unmanage it and subscribe to property changes (this always worked). In recent xmonad-contrib, however, both values should be fine, and thus `overrideRedirect = True` (the default) should be preferred as it then also lowers on start by default.

We could possibly add `doLower` to the manage hook in ManageDocks to make both options of `overrideRedirect` just work. I'm a bit afraid of that, though, assuming there must be a reason we haven't done that years ago. :-)

PR #281, opened 2021-03-23
Merge X.L.LayoutCombinators.(|||) into X.L.(|||)

slotThe:
### Description

The functionality of the forme are quite handy to have in core and we
can do so with minimal code changes.

The drawbacks of this approach are

  1. We can't merge `JumpToLayout` into `ChangeLayout` because people may
     have imported `JumpToLayout(..)`, which we can't simulate with type
     aliases

  2. Because the internal structure of `X.L.LayoutCombintors.(|||)` is a
     bit different, this creates a regression for people who used
     `NextLayoutNoWrap` or `Wrap` in their configs.  We could work around
     this by creating fake instances of these fields in the new
     `JumpToLayout` constructor and simply not doing anything with them,
     but since this seems like quite an advanced and specific use-case,
     failing fast and hard (as opposed to adding deprecation messages
     and then "silently" not handling these messages) seems preferrable
     here.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -44,6 +44,10 @@
   * Added the `extensibleConf` field to `XConfig` which makes it easier for
     contrib modules to have composable configuration (custom hooks, …).
 
+  * Migrated `X.L.LayoutCombinators.(|||)` into `XMonad.Layout`,
+    providing the ability to directly jump to a layout with the
+    `JumpToLayout` message.
+
 ## 0.15 (September 30, 2018)
 
   * Reimplement `sendMessage` to deal properly with windowset changes made
--- a/src/XMonad/Layout.hs
+++ b/src/XMonad/Layout.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE FlexibleInstances, MultiParamTypeClasses, PatternGuards, TypeSynonymInstances, DeriveDataTypeable #-}
+{-# LANGUAGE FlexibleInstances, MultiParamTypeClasses, PatternGuards, TypeSynonymInstances, DeriveDataTypeable, LambdaCase #-}
 
 -- --------------------------------------------------------------------------
 -- |
@@ -16,7 +16,7 @@
 
 module XMonad.Layout (
     Full(..), Tall(..), Mirror(..),
-    Resize(..), IncMasterN(..), Choose(..), (|||), CLR(..), ChangeLayout(..),
+    Resize(..), IncMasterN(..), Choose(..), (|||), CLR(..), ChangeLayout(..), JumpToLayout(..),
     mirrorRect, splitVertically,
     splitHorizontally, splitHorizontallyBy, splitVerticallyBy,
 
@@ -30,7 +30,7 @@ import Graphics.X11 (Rectangle(..))
 import qualified XMonad.StackSet as W
 import Control.Arrow ((***), second)
 import Control.Monad
-import Data.Maybe (fromMaybe)
+import Data.Maybe (fromMaybe, isNothing)
 
 ------------------------------------------------------------------------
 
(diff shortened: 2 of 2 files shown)


geekosaur:
Speling, for what it's worth :)

slotThe:
Whoops :)

CHANGES.md
```diff
@@ -35,6 +35,10 @@
   * Fixed dunst notifications being obscured when moving floats.
     https://github.com/xmonad/xmonad/issues/208
 
+  * Migrated `X.L.LayoutCombinator.(|||)` into `XMonad.Layout`,
```
geekosaur: `LayoutCombinators`

geekosaur (approved):
LGTM.

liskin (approved):
This is way less messy than I expected. 👍
Sorry for the delay.

Does this need https://github.com/xmonad/xmonad-contrib/pull/493/ to be merged to avoid breakage or is that PR optional? Asking just to understand the dependencies and migration path.

liskin:
Anyway, please mention https://github.com/xmonad/xmonad-contrib/pull/493 in the commit message :-)

slotThe:
> Does this need [xmonad/xmonad-contrib#493](https://github.com/xmonad/xmonad-contrib/pull/493) to be merged to avoid breakage or is that PR optional? Asking just to understand the dependencies and migration path.

Yes, contrib doesn't compile with this change and without xmonad/xmonad-contrib#493; X.L.LayoutCombinators also defines a `JumpToLayout` constructor and then GHC can't figure out which one we mean.  

I'll rebase and mention that pr in the commit message

liskin:
> Yes, contrib doesn't compile with this change and without [xmonad/xmonad-contrib#493](https://github.com/xmonad/xmonad-contrib/pull/493); X.L.LayoutCombinators also defines a `JumpToLayout` constructor and then GHC can't figure out which one we mean.

Oh, looks like top-level definitions are in the same lexical scope as imported stuff, I didn't know that.

In that case I'd like to release X11 first, and only then do a "big bang" where we bump versions and dep bounds and merge all the build breaking stuff.

src/XMonad/Layout.hs
```diff
@@ -155,6 +190,18 @@ instance Message NextNoWrap
 handle :: (LayoutClass l a, Message m) => l a -> m -> X (Maybe (l a))
 handle l m = handleMessage l (SomeMessage m)
 
+-- | Swap the two sub-layouts.
```
liskin: This is misleading. I'd expect a `Choose l r a → Choose r l a`, but instead it's a `switch` that also happens to notify the inner layouts that they're going to be hidden.

liskin: Oh and I just noticed that there's this `choose` function that could possibly replace both `swap` and `switch`, as it seems to be more general. IOW, `JumpToLayout` handling may be possible to implement concisely using `choose` alone.

liskin (changes requested):
Let's try to simplify the impl and merge asap to have at least a few weeks of testing.

slotThe:
> Let's try to simplify the impl and merge asap to have at least a few weeks of testing.

Sure, I'll give that a try tomorrow

slotThe:
Turns out that you were right!  I have pushed 246600f70d6365c612ea9bfd31efd50c50a6c861, which works for me (so fingers crossed it does for others as well :))

src/XMonad/Layout.hs
```diff
@@ -247,22 +230,19 @@ instance (LayoutClass l a, LayoutClass r a) => LayoutClass (Choose l r) a where
...
-             if isNothing m'
-                 then traverse (swap . switch) =<< passOnM m (switch c)
-                 else return m'
+    handleMessage c@(Choose d l r) m | Just (JumpToLayout desc) <- fromMessage m =
+        case d of
+            CL -> if   desc == description r
+                  then choose c CR Nothing Nothing
+                  else handleMessage l m >>= \case
+                      Just ml -> choose c CL (Just ml) Nothing
+                      Nothing -> choose (Choose CR l r) CR Nothing =<< handleMessage r m
```
liskin: Why do you manually construct `(Choose CR l r)` instead of using `c` here? I think that will suppress the `Hide` events.

slotThe: `c` is of the form `Choose CL l r`, if I wanted to use it I'd need another `switch`-like function.  Putting `c` there would hide the current layout in every case, even if we're on, say, `Full` and executed `sendMessage (JumpToLayout Full)`

liskin: If we're on `Full` and we `JumpToLayout Full`, it's handled by the `desc == description r` case above, isn't it?

`choose (Choose CR l r) CR` and `choose (Choose CL l r) CL` will never `hide` anything, and that doesn't sound right. The `l` layout does need hiding when we're switching away from it, I think.

src/XMonad/Layout.hs
liskin: (If you want to test this, I expect that any decorated layout would fail to destroy its decoration windows if Hide messages aren't delivered to it on time.)

src/XMonad/Layout.hs
slotThe: This all makes perfect sense, yet it still works (even with decorated layouts)...  I'm very confused.

We can, however, at least cover the "we're on the same layout" part by checking for descriptions in either case.  This, like magic, also enables us to get rid of the non-hiding calls to `choose` ( 0084a1af5c08fa3015af5aa2fd45d0c165d86a67 )

liskin: Can you explain what the problem was with the "we're on the same layout" part? That may have been the bit I missed…

slotThe: > Can you explain what the problem was with the "we're on the same layout" part? That may have been the bit I missed…

You said that 

  > If we're on `Full` and we `JumpToLayout Full`, it's handled by the `desc == description r` case above, isn't it?

but it might well be that our layouts looks like `Choose CL Full (*other-layouts*)` and so we'd have to check `desc == description l` in the `CL` case as well (which we and also the old `X.L.LayoutCombinators.(|||)` didn't do).  As it stood, this would have then gone through to the other cases, which then managed to switch the layout (I imagine `handleMessage l m` returned `Nothing` and then we flipped `CL` to `CR` in `choose`)

src/XMonad/Layout.hs
```diff
@@ -230,16 +230,13 @@ instance (LayoutClass l a, LayoutClass r a) => LayoutClass (Choose l r) a where
...
-           | otherwise             -> case d of
-               CL -> handleMessage l m >>= \case
-                   Just ml -> choose c CL (Just ml) Nothing
-                   Nothing -> choose c CR Nothing =<< handleMessage r m
-               CR -> handleMessage r m >>= \case
-                   Just mr -> choose c CR Nothing (Just mr)
-                   Nothing -> (\ml -> choose c CL ml Nothing) =<< handleMessage l m
+    handleMessage c@(Choose d l r) m | Just (JumpToLayout desc) <- fromMessage m = do
+        ml <- handleMessage l m
+        mr <- handleMessage r m
```
liskin, line 235: I should probably not even mention it to you, as we all know what's gonna happen, but still:

This simple implementation always walks the whole tree of Chooses and then possibly passes the message all the way down layout modifiers as well. The previous one stopped descending when it found a node with the desired description, but it still passed the message to some subtrees (all the way to layout modifiers and base layout). Asymptotically it's the same, in practice it might be a couple nanosecs or microsecs difference, I wouldn't worry about it at all. But it can be dealt with by keeping the first two lines of the multi-way if in here.

slotThe: I will go against expectations and prefer clarity to this spurious efficiency gain in this situation then :P

PR #284, opened 2021-03-28
Broadcast `DestroyWindowEvent` to layouts

TheMC47:
### Description

Some layout and layout modifiers that keep track of some window
properties don't do garbage collection and update their state when
windows are destroyed. By broadcasting this event, it should be easier
for layouts to clean up

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -50,6 +50,9 @@
 
   * Document the help command in the help message.
 
+  * `DestroyWindowEvent` is now broadcasted to layouts to let them know
+    window-specific resources can be discarded.
+
 ## 0.15 (September 30, 2018)
 
   * Reimplement `sendMessage` to deal properly with windowset changes made
--- a/src/XMonad/Layout.hs
+++ b/src/XMonad/Layout.hs
@@ -27,6 +27,7 @@ module XMonad.Layout (
 import XMonad.Core
 
 import Graphics.X11 (Rectangle(..))
+import Graphics.X11.Xlib.Extras ( Event(DestroyWindowEvent) )
 import qualified XMonad.StackSet as W
 import Control.Arrow ((***), second)
 import Control.Monad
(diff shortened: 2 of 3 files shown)


liskin:
I see one potential problem with this: if we want this to really be reliable, we'll need to treat `DestroyWindowEvent` the same special way as `ReleaseResources`:
https://github.com/xmonad/xmonad/blob/5cdf428f5513e14f7e13c707ceeca1ed42207625/src/XMonad/Layout.hs#L201
https://github.com/xmonad/xmonad-contrib/blob/902c2bb17d1a1ee4a56944dbf952d6f95d5700c6/XMonad/Layout/LayoutCombinators.hs#L264
i.e. send it to inactice layouts as well.

I'm not sure if it's worth it. It very well can be, I just don't know at this point, so I intend to revisit this later. Other opinions welcome!

geekosaur:
I think it's worth it, and we can mostly track where by looking for `ReleaseResources` already.

One potential issue we're not handling currently is reuse of window `XID`s, which may lead to odd behavior in any layout or modifier which tracks them. (This may include fullscreening or `WindowSwallowing`, etc.)

liskin:
Well then, let's go ahead with this.
@TheMC47 can you please broadcast the message to inactive layouts? Then we can merge it. If we also merge https://github.com/xmonad/xmonad/pull/281 soon enough, we probably don't even need to worry about xmonad-contrib that much (just obscure stuff like SubLayouts and I can take care of that later).

TheMC47:
> @TheMC47 can you please broadcast the message to inactive layouts?

Not quite sure how to do that exactly

liskin:
> Not quite sure how to do that exactly

grep for ReleaseResources and treat DestroyWindowEvent similarly in the Choose layout (https://github.com/xmonad/xmonad/pull/284#issuecomment-812161492)

TheMC47:
@liskin except for rebasing, is there anything that needs to be done here?

liskin:
@TheMC47 nothing that I'm aware of, just waits till I refocus onto merging core xmonad PRs again

(I'll be in vacation mode till 2021-07-12, meaning I can talk but I probably won't really focus on coding)

liskin (approved):
Pushed a minor fixup, will merge soon, ping me when you've seen it.

TheMC47:
@liskin all good, thanks!

PR #267, opened 2021-01-20
Revert "Re-enable lts based testing"

peti:
Reverts xmonad/xmonad#266. This patch loses the ability to build `generatemanpage`.

PR #285, opened 2021-03-31
GenerateManpage: Simplify and invoke in CI (variant A: simplify)

### LINKED ISSUE #283: xmonad: consolidate #268 and #260
In a zoom call with @peti, we agreed that:

* don't: rely on undeclared deps like the pandoc binary in released xmonad tarball
* do: rebuild manpage in CI
* don't: ship both generated manpage and the script to generate it in the tarball (will stay in git repo only)
* unnecessary: use a specific version of pandoc that isn't widely available outside of cabal
* acceptable: simplify generatemanpage to make it easier to rebuild it in CI and/or dev machines

Plan:

* drop generatemanpage from cabal, and then either
  * add a separate tiny cabal project for the script and its deps, or
  * simplify it enough that it doesn't need any cabal project (uses only libs shipped with ghc, or is just a sed script or something)
* rebuild manpage on push to master (#260-ish)
* restore testing against more LTS versions with stack (#268, plus maybe some additional bits from https://github.com/xmonad/X11/pull/75)

liskin:
Fixes https://github.com/xmonad/xmonad/issues/283.
See https://github.com/xmonad/xmonad/pull/286 for variant B: cabalise.

#### [GenerateManpage: Drop all dependencies except base and invoke in CI](../commit/0f677b22d3da65dd34ecbc8bdf3f091640b1e87b)

pandoc's API changes often enough that distros like Debian were patching our GenerateManpage.hs to work with their version of pandoc, and it doesn't build against any Stackage LTS except the recently released LTS-17. Also, building pandoc from source takes quite some time and resources.

But for what benefit? We're not using any special pandoc functionality whatsoever. It's just that it was all in Haskell and the entire build was orchestrated by cabal, but is all that complexity and resource consumption worth it? I think not.

(Frankly, this whole thing was just a massive waste of time as the help text in Config.hs isn't generated at all, so we still need to do this manually. And then, the default key-bindings in core are unlikely to change ever again.)

Let's simplify this:

* drop all dependencies except base and just run it through runhaskell

* add a Makefile and GH Actions workflow that invokes this after push

* only ship the results in release tarball, not the scripts which are
  considered our dev infrastructure

Closes: https://github.com/xmonad/xmonad/issues/283
Related: https://github.com/xmonad/xmonad/pull/260
Related: https://github.com/xmonad/xmonad/pull/261
Related: https://github.com/xmonad/xmonad/pull/265
Related: https://github.com/xmonad/xmonad/pull/266
Related: https://github.com/xmonad/xmonad/pull/267
Related: https://github.com/xmonad/xmonad/pull/268

### Checklist


liskin:
@psibi @peti Can you please take a look at this and #286? Thanks!

CHANGES.md
```diff
@@ -35,6 +35,10 @@
   * Fixed dunst notifications being obscured when moving floats.
     https://github.com/xmonad/xmonad/issues/208
 
+  * `util/GenerateManpage.hs` is no longer distributed in the tarball.
+    Instead, the manpage source is renegerated and manpage rebuilt
```
psibi:
```suggestion
    Instead, the manpage source is regenerated and manpage rebuilt
```

psibi (approved):

CHANGES.md
liskin: Oh, stupid vim treats indented bullets as code blocks and refuses to spellcheck in there. :-(

slotThe:
@peti gentle ping for this and #286 :)

slotThe (approved):
I think that at this point enough time has passed for people to give their input.  

Personally, I think we should go with this over #286.  To prevent `0.16` (and thusly also contrib `0.17`) from stalling, I would push for a decision this coming weekend

PR #286, opened 2021-03-31
GenerateManpage: Clean up and invoke in CI (variant B: cabalise)

### LINKED ISSUE #283: xmonad: consolidate #268 and #260

liskin:
Fixes: https://github.com/xmonad/xmonad/issues/283
See https://github.com/xmonad/xmonad/pull/285 for variant A: simplify.

#### [GenerateManpage: Factor into separate cabal project, invoke in CI](../commit/52409ec1c35e7eb744656fc2bda15c312f81915f)

pandoc's API changes often enough that distros like Debian were patching our GenerateManpage.hs to work with their version of pandoc, and it doesn't build against any Stackage LTS except the recently released LTS-17. Also, building pandoc from source takes quite some time and resources.

Therefore, separate it into its own cabal project and invoke it in CI to make sure we're releasing an up-to-date manpage, and don't encourage downstreams to rebuild it -- it's now just our dev tooling.

(Uncached run of the generatemanpage workflow takes more than half an hour, which seems incredibly wasteful for what it does, and that's the reason why I add GenerateManpage.cabal.project.freeze. The default key-bindings in core are unlikely to change ever again, the help text in Config.hs isn't generated at all, so we still need to do that thing manually, I'm really not convinced this is worth the trouble...)

Closes: https://github.com/xmonad/xmonad/issues/283
Related: https://github.com/xmonad/xmonad/pull/260
Related: https://github.com/xmonad/xmonad/pull/261
Related: https://github.com/xmonad/xmonad/pull/265
Related: https://github.com/xmonad/xmonad/pull/266
Related: https://github.com/xmonad/xmonad/pull/267
Related: https://github.com/xmonad/xmonad/pull/268

#### [GenerateManpage: Write markdown as well](../commit/028b81510009c44d00687e133a276142e4b913f7)

Now that we only ship the generated outputs in the tarball, let's have the generated keybindings in the xmonad.1.markdown.

#### [GenerateManpage: Reintroduce version number to manpage](../commit/30a8f9272b87b1dce893bf97aabbedd83b25cf30)

Dropped in 886a0d4041420b4d653c6aed51bb7c780991d76a, but easy to put back, assuming Cabal API stays stable for a while. :-)

### Checklist


liskin:
@psibi @peti Can you please take a look at this and #285? Thanks!

liskin:
Closed in favor of https://github.com/xmonad/xmonad/pull/285

PR #288, opened 2021-04-01
Add `MINIMAL` for `ExtensionClass` and `LayoutClass`

slotThe:
### Description

Add a `MINIMAL` pragma for the type classes that should really have had one in the first place (i.e. the ones that currently have some kind of "minimal complete definition" in their docstring).  Since we now only support GHC versions that support that kind of pragma, it seems like the right thing to do.  I left the comments in for now, but I suppose an argument could be made that they should be removed, since Haddock will correctly generate a `Minimal complete definition` heading.

I also added the relevant definitions for `Full` and `Tall`, so there's no scary warnings generated.

### Checklist

  ~~- [ ] I updated the `CHANGES.md` file~~

### DIFF
--- a/src/XMonad/Core.hs
+++ b/src/XMonad/Core.hs
@@ -387,6 +387,7 @@ instance Message LayoutMessages
 --
 -- Minimal complete definition: initialValue
 class Typeable a => ExtensionClass a where
+    {-# MINIMAL initialValue #-}
     -- | Defines an initial value for the state extension
     initialValue :: a
     -- | Specifies whether the state extension should be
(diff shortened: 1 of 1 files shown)


geekosaur (approved):
LGTM!

src/XMonad/Core.hs
```diff
@@ -338,6 +338,9 @@ class (Show (layout a), Typeable layout) => LayoutClass layout a where
     description :: layout a -> String
     description      = show
 
+    {-# MINIMAL (runLayout | ((doLayout | pureLayout) , emptyLayout))
```
liskin: Did you test this with xmonad-contrib? Few layouts override emptyLayout, so this will result in lots of warnings.

It seems to me the minimal complete definition of a LayoutClass is the one that overrides nothing, as `instance LayoutClass Full` shows, and that it is completely fine that way.

slotThe: Mh you're right.  I guess we should change the comment then?  Something along the lines of highlighting the dependencies of the different functions?

slotThe: Okay I pushed a doc change instead

PR #291, opened 2021-04-13
added export list to Operations, refactored message handling.

ghost:
### Description

The current definition of `broadcastMessage` seems to be O(n^2) in the number of workspaces because it uses `sendMessageWithNoRefresh` and `sendMessageWithNoRefresh` uses `updateLayout` and `updateLayout` uses `runOnWorkspaces`.

This changes `broadCastMessage` and `sendMessageWithNoRefresh` to each use a single call to `runOnWorkspaces`.

Because it uses two new helper functions, I also added an explicit export list to `XMonad.Operations`. The export list defines sections as indicated by comments in the source code. (`XMonad.Operations` could use a documentation cleanup, but this is not that.)

### Checklist

### DIFF
--- a/src/XMonad/Operations.hs
+++ b/src/XMonad/Operations.hs
@@ -14,7 +14,41 @@
 --
 -----------------------------------------------------------------------------
 
-module XMonad.Operations where
+module XMonad.Operations (
+    -- * Manage One Window
+    manage, unmanage, killWindow, kill, isClient,
+    setInitialProperties, setWMState, setWindowBorderWithFallback,
+    hide, reveal, tileWindow,
+    setTopFocus, focus, withFocused,
+
+    -- * Manage Windows
+    windows, refresh, rescreen, modifyWindowSet, windowBracket, windowBracket_, clearEvents, getCleanedScreenInfo,
+
+    -- * Keyboard and Mouse
+    cleanMask, extraModifiers,
+    mouseDrag, mouseMoveWindow, mouseResizeWindow,
+    setButtonGrab, setFocusX,
+
+    -- * Messages
+    sendMessage, broadcastMessage, sendMessageWithNoRefresh,
+
+    -- * Save and Restore State
+    StateFile (..), writeStateToFile, readStateFile, restart,
+
+    -- * Floating Layer
+    float, floatLocation,
+
+    -- * Window Size Hints
+    D, mkAdjust, applySizeHints, applySizeHints', applySizeHintsContents,
+    applyAspectHint, applyResizeIncHint, applyMaxSizeHint,
+
+    -- * Rectangles
+    containedIn, nubScreens, pointWithin, scaleRationalRect,
+
+    -- * Other Utilities
+    initColor, pointScreen, screenWorkspace,
+    setLayout, updateLayout,
+    ) where
 
 import XMonad.Core
 import XMonad.Layout (Full(..))
--- a/src/XMonad/Operations.hs
+++ b/src/XMonad/Operations.hs
@@ -455,8 +455,7 @@ filterMessageWithNoRefresh p a = updateLayoutsBy $ \ wrk ->
       else pure Nothing
 
 -- | Update the layouts of some workspaces.
-updateLayoutsBy :: MonadState XState m
-                => (WindowSpace -> m (Maybe (Layout Window))) -> m ()
+updateLayoutsBy :: (WindowSpace -> X (Maybe (Layout Window))) -> X ()
 updateLayoutsBy f = runOnWorkspaces' $ \ wrk ->
     maybe wrk (\ l' -> wrk{ W.layout = l' }) <$> f wrk
 
--- a/src/XMonad/Operations.hs
+++ b/src/XMonad/Operations.hs
@@ -63,6 +63,7 @@ import Data.Ratio
 import qualified Data.Map as M
 import qualified Data.Set as S
 
+import Control.Applicative (liftA3)
 import Control.Arrow (second)
 import Control.Monad.Reader
 import Control.Monad.State
@@ -456,13 +457,13 @@ filterMessageWithNoRefresh p a = updateLayoutsBy $ \ wrk ->
 
 -- | Update the layouts of some workspaces.
 updateLayoutsBy :: (WindowSpace -> X (Maybe (Layout Window))) -> X ()
-updateLayoutsBy f = runOnWorkspaces $ \ wrk ->
+updateLayoutsBy f = runOnWorkspaces' $ \ wrk ->
     maybe wrk (\ l' -> wrk{ W.layout = l' }) <$> f wrk
 
 -- | Update the layout field of a workspace.
 updateLayout :: WorkspaceId -> Maybe (Layout Window) -> X ()
 updateLayout i ml = whenJust ml $ \l ->
-    runOnWorkspaces $ \ww -> return $ if W.tag ww == i then ww { W.layout = l} else ww
+    runOnWorkspaces' $ \ww -> return $ if W.tag ww == i then ww { W.layout = l} else ww
 
 -- | Set the layout of the currently viewed workspace.
 setLayout :: Layout Window -> X ()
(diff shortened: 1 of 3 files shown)


src/XMonad/Operations.hs
```diff
@@ -403,17 +430,24 @@ sendMessage a = windowBracket_ $ do
...
 broadcastMessage :: Message a => a -> X ()
-broadcastMessage a = withWindowSet $ \ws -> do
-   let c = W.workspace . W.current $ ws
-       v = map W.workspace . W.visible $ ws
-       h = W.hidden ws
-   mapM_ (sendMessageWithNoRefresh a) (c : v ++ h)
+broadcastMessage = filterMessageWithNoRefresh (const True)
 
 -- | Send a message to a layout, without refreshing.
 sendMessageWithNoRefresh :: Message a => a -> W.Workspace WorkspaceId (Layout Window) Window -> X ()
```
slotThe: This may be a good time to change this to `Message a => a -> WindowSpace -> X ()` for more clarity.

liskin:
This looks nice, but there's one concern I have: `runOnWorkspaces` runs the monadic action on the hidden workspaces first, then current and visible, whereas `broadcastMessage` did it in current, visible, hidden order. I'm not entirely sure it's safe to change this order, as some may depend on it. :-/
(Opinions from others welcome.)

Other than that, the accompanying text that you wrote in the pull request description should go into the commit message as well. Also, whenever you feel like using comma in the commit message subject, that's a hint you might consider splitting it into two. ;-)
More at https://chris.beams.io/posts/git-commit/, https://who-t.blogspot.com/2009/12/on-commit-messages.html

ghost:
> Also, whenever you feel like using comma in the commit message subject, that's a hint you might consider splitting it into two. ;-)
> More at https://chris.beams.io/posts/git-commit/, https://who-t.blogspot.com/2009/12/on-commit-messages.html

Good call. I'll split this up into 3 PRs: one for the export list, one to cleanup documentation comments, and one to do the actual refactor I want.

liskin:
> Good call. I'll split this up into 3 PRs: one for the export list, one to cleanup documentation comments, and one to do the actual refactor I want.

Oh no please don't. Commits, not PRs! Please reopen this one, there's absolutely no need to add more noise with additional PRs.

ghost:
> > Good call. I'll split this up into 3 PRs: one for the export list, one to cleanup documentation comments, and one to do the actual refactor I want.
> 
> Oh no please don't. Commits, not PRs! Please reopen this one, there's absolutely no need to add more noise with additional PRs.

Got it. Sorry about the confusion! Three separate commits now. Should be clear what they do. This goes a bit farther than the previous commits because it fixes up the documentation a bit too. Things looked really odd with correct sections but incorrect Haddock comments.

Now I just need to deal with order-of-operations issues. For now I'm thinking of writing a version of `runOnWorkspaces` that uses the same order as the original `broadcastMessage`.

ghost:
OK, the order of operations is as before. To do this I added two more non-exported functions. One traverses the workspaces in a stackset in the correct order, the other one does `runOnWorkspaces` in the right order. Both might be useful to export from somewhere (probably XMonad.StackSet for the traversal), but that's probably out of scope for this PR.

src/XMonad/Operations.hs
```diff
@@ -190,8 +191,9 @@ windowBracket p action = withWindowSet $ \old -> do
     windows         $ \_ -> new
   return a
 
--- | A version of @windowBracket@ that discards the return value, and handles an
--- @X@ action reporting its need for refresh via @Any@.
+-- | Perform an @X@ action. If it returns @Any True@, unwind the changes to the @WindowSet@ and replay them using @windows@.
+-- This is a version of @windowBracket@ that discards the return value and
+-- handles an @X@ action that reports its need for refresh via @Any@.
```
slotThe: I think it's worth trying to be consistent with the doc-style of the rest of the module here (and also in other parts); meaning line-breaks after what I assume are 71 or 79 columns.

src/XMonad/Operations.hs
```diff
@@ -14,7 +14,33 @@
 --
 -----------------------------------------------------------------------------
 
-module XMonad.Operations where
+module XMonad.Operations (
```
slotThe, line 16: This is really cool :+1:

src/XMonad/Operations.hs
```diff
@@ -431,17 +431,24 @@ sendMessage a = windowBracket_ $ do
...
-       h = W.hidden ws
-   mapM_ (sendMessageWithNoRefresh a) (c : v ++ h)
+broadcastMessage = filterMessageWithNoRefresh (const True)
 
 -- | Send a message to a layout, without refreshing.
 sendMessageWithNoRefresh :: Message a => a -> W.Workspace WorkspaceId (Layout Window) Window -> X ()
 sendMessageWithNoRefresh a w =
-    handleMessage (W.layout w) (SomeMessage a) `catchX` return Nothing >>=
-    updateLayout  (W.tag w)
+    filterMessageWithNoRefresh (on (==) W.tag w) a
```
slotThe: I don't know if this is just me, but I always find `on` to be terribly confusing if it's not written in an infix style, because when it reads like "do something on `(==)`" instead of "do `(==)` on something`

src/XMonad/Operations.hs
```diff
@@ -496,6 +497,31 @@ initColor dpy c = C.handle (\(C.SomeException _) -> return Nothing) $
...
+  where
+    windowsetM_ f = do
+        ws' <- f =<< gets windowset
+        modify $ \ s -> s{ windowset = ws' }
+
+-- | Traverse an 'Applicative' function over the workspaces, starting with the current.
+workspacesA ::
+    (Applicative m)=>
+    (W.Workspace i l a -> m (W.Workspace i' l' a)) ->
+    W.StackSet i l a sid sd -> m (W.StackSet i' l' a sid sd)
```
slotThe: This formatting (ditto `runOnWorkspaces'`) looks kind of funny

liskin:
Just got back to this and I noticed `runOnWorkspaces` isn't used anywhere else than in `updateLayout`, so it might be sensible to just change it instead of introducing `runOnWorkspaces'`. Nothing in xmonad-contrib uses it either, so we don't need to think about whether something depends on the order or not—it's simply not used at all. Possibly a very small number of users have this in their config, but a changelog entry will deal with this.

Sorry for the long delay. No need to rush making the change, either, I'll be happy to take over and finish it and merge later next week.

src/XMonad/Operations.hs
ghost: OK, switched to infix `on`. I'm used to thinking of infix vs. prefix as a way to save on parentheses (in this case there are fewer prefix). But I see how, as a word, using it the 'wrong' way seems nongrammatical.

ghost:
No worries about the delay. I agree that there's absolutely no rush. It's a change that, hopefully, does not change anything except for speeding up message handling when you have a lot of workspaces.

Please proceed with as much caution as you think is necessary. I haven't run into any problems, but better to have consistent n^2 message handling than someone's complex setup breaking because messages arrive at layouts in a slightly different order.

I've cleaned up the formatting per @slotThe 's helpful suggestions and rebased.

src/XMonad/Operations.hs
ghost: Feel free to change it if you have a better formatting (do you need / can I give permission?). For long signatures like this I do my best to wrap them in semantic units that fit in a Tall layout on my little laptop screen. With an `Applicative` constraint, give a function from a workspace to an action on a workspace to get a function from a stackset to an action on a stackset.

src/XMonad/Operations.hs
ghost: That's embarrassing. Fixed now.

src/XMonad/Operations.hs
geekosaur: Once you convert a branch on your repo into a pull request, any xmonad maintainer has access to the PR. Of course, you're asking for trouble with force pushes in that case.

slotThe: You can tick the "allow access from maintainers" checkbox somewhere in the PR, but I think it is on by default.  I think the grouping is quite understandable; I was just thrown off a bit due to the trailing arrow syntax, as opposed to something like 

``` haskell
workspacesA
    :: Applicative m
    => (W.Workspace i l a -> m (W.Workspace i' l' a))
    -> W.StackSet i l a sid sd -> m (W.StackSet i' l' a sid sd)
```

ghost: Ironically the prefix form is confusing to me because the type symbol `::` is on the constraint, the constraint arrow `=>` is on a parameter, and a parameter arrow `->` is on the result. But that does seem to be the norm for XMonad, so I'll switch it over.

Should I stop doing force pushes on this PR?

liskin:
> Please proceed with as much caution as you think is necessary.

Just a heads up what my plan here is: I'll change the `runOnWorkspaces` order and thus drop some of the new functions in X.Core. I'm quite confident it's safe enough to do. I meant to do that yesterday but spent most of the day packing my backpack instead. I'll get back to this next wednesday/thursday. No action needed, I'll finish this.

liskin:
Not exactly "next wednesday/thursday" but it's done! Thanks. :-)

PR #292, opened 2021-04-13
Add export list to Operations

ghost:
An export list will allow folks to add functions to the module without adding them to the API, and allow Haddock to divide the documentation into sections.

The API of the module is not changed. (Every function is still exported, it's now just explicit.)

The Haddock sections will make the module documentation render differently.

### Checklist


liskin:
Please update #291 instead.

PR #294, opened 2021-05-17
Add support for extensible config in contrib modules

liskin:
### Description

It's often difficult to make contrib modules work together. When one depends on a functionality of another, it is often necessary to expose lots of low-level functions and hooks and have the user combine these into a complex configuration that works. This is error-prone, and arguably a bad UX in general.

This commit presents a simple solution to that problem inspired by "extensible state": extensible config. It allows contrib modules to store custom configuration values inside XConfig. This lets them create custom hooks, ensure they hook into xmonad core only once, and possibly other use cases I haven't thought of yet.

For more, see the related pull request to xmonad-contrib.

Related: https://github.com/xmonad/xmonad-contrib/pull/547

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -41,6 +41,9 @@
     Instead, the manpage source is regenerated and manpage rebuilt
     automatically in CI.
 
+  * Added the `extensibleConf` field to `XConfig` which makes it easier for
+    contrib modules to have composable configuration (custom hooks, …).
+
 ## 0.15 (September 30, 2018)
 
   * Reimplement `sendMessage` to deal properly with windowset changes made
--- a/src/XMonad/Config.hs
+++ b/src/XMonad/Config.hs
@@ -277,6 +277,7 @@ instance (a ~ Choose Tall (Choose (Mirror Tall) Full)) => Default (XConfig a) wh
     , XMonad.handleExtraArgs = \ xs theConf -> case xs of
                 [] -> return theConf
                 _ -> fail ("unrecognized flags:" ++ show xs)
+    , XMonad.extensibleConf     = M.empty
     }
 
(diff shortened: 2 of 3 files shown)


ghost:
Should `extensibleConf` be a map from `Fingerprint`, not `TypeRep`?

`Fingerprint` seems practically designed for this use case. Although read-only state doesn't have to be serialized, it would be nice to use the same key type if `extensibleState` ever makes the switch.

liskin:
This [`Fingerprint`](https://hackage.haskell.org/package/base-4.15.0.0/docs/GHC-Fingerprint.html)? What benefits does it have over `TypeRep`? Its docs doesn't mention anything type-related either. Did I miss something?

ghost:
I just being dumb. I was looking at [`typeRepFingerprint`](https://hackage.haskell.org/package/base-4.15.0.0/docs/Data-Typeable.html#g:7) ("Takes a value of type a and returns a concrete representation of that type"), but the documentation seems inaccurate and `Fingerprint` isn't exported from Typeable (need to use GHC.Fingerprint.Type).

So it looks like there's a type designed to act as a key for types, but in its current state it's not especially usable. Sorry about that.

liskin:
Oh, that's probably the same Fingerprint I linked. So yeah, we could possibly use that, and for `extensibleState` that might be a bit faster than a Map keyed by String. But we'd be relying on GHC internals that aren't really documented in the public API, notably the fingerprints being injective (fpr equality implies type equality), and also being stable across recompiles and version changes and so on. I don't think that's worth it at the moment. If extensible state/conf was a measurable performance bottleneck, then sure, but it looks as just a premature optimization at this point.

PR #297, opened 2021-05-18
Prevent grabKey from accidentally grabbing all unbound keys

### LINKED ISSUE #293: `bindKeys` binds `KeySym` 0 to all unbound keys
`bindKeys` doesn't check that a `KeySym` is valid before looking up the keycode(s) it's bound to. In particular, `KeySym` 0 gets mapped to every unbound keycode, since they will have `NoSymbol` == 0 for their base `KeySym`.

This can most easily be reproduced using `statusBar` in xmonad-contrib, with `def` as the key-mapping function; this unexpectedly invokes the following instances:

```haskell
instance Default b => a -> b where def = const def
instance (Default a, Default b) => (a, b) where def = (def, def)
instance Default CInt where def = 0
```
thus producing a function which binds `toggleStruts` to the `(KeyMask,KeySym)` pair `(0,0)` (and demonstrating why `Default` is a dangerous abstraction). Note that it is not reliably possible to override `def` in this situation; the only thing we can do is try to avoid the aftermath, and I would expect that an inadvertent 0 `KeySym` is the most common error here anyway.

### Checklist

  - [x] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

liskin:
### Description

grabKeys doesn't check that a KeySym is valid before looking up the KeyCode(s) it's bound to. In particular, KeySym 0 (NoSymbol) gets mapped to every unbound KeyCode, since that's what XKeycodeToKeysym returns for those.

This can most easily be reproduced using `statusBar` in xmonad-contrib, with def as the key-mapping function; this unexpectedly invokes the following instances:

    instance Default b => a -> b where def = const def
    instance (Default a, Default b) => (a, b) where def = (def, def)
    instance Default CInt where def = 0

thus producing a function which binds `toggleStruts` to the (KeyMask, KeySym) pair (0, 0) (and demonstrating why Default is a dangerous abstraction). Note that it is not reliably possible to override `def` in this situation; the only thing we can do is try to avoid the aftermath, and I would expect that an inadvertent 0 KeySym is the most common error here anyway.

Fixes: https://github.com/xmonad/xmonad/issues/293
Fixes: 40cb12ce174a ("Grab all keycodes linked to each keysym, not just one")
Co-authored-by: @geekosaur

### Checklist

    Tested locally using `io $ print keysymMap`

    I believe this is so minor that it's just noise to most people.

### DIFF
--- a/src/XMonad/Main.hs
+++ b/src/XMonad/Main.hs
@@ -473,8 +473,11 @@ grabKeys = do
     -- build a map from keysyms to lists of keysyms (doing what
     -- XGetKeyboardMapping would do if the X11 package bound it)
     syms <- forM allCodes $ \code -> io (keycodeToKeysym dpy code 0)
-    let keysymMap = M.fromListWith (++) (zip syms [[code] | code <- allCodes])
-        keysymToKeycodes sym = M.findWithDefault [] sym keysymMap
+    let keysymMap' = M.fromListWith (++) (zip syms [[code] | code <- allCodes])
+    -- keycodeToKeysym returns noSymbol for all unbound keycodes, and we don't
+    -- want to grab those whenever someone accidentally uses def :: KeySym
+    let keysymMap = M.delete noSymbol keysymMap'
+    let keysymToKeycodes sym = M.findWithDefault [] sym keysymMap
     forM_ (M.keys ks) $ \(mask,sym) ->
          forM_ (keysymToKeycodes sym) $ \kc ->
               mapM_ (grab kc . (mask .|.)) =<< extraModifiers


liskin:
@geekosaur There's one thing that still isn't clear to me: what bad thing happens when we grab all the unbound keycodes? I don't think I have any actual keys on my keyboard that aren't bound. Is that an issue only for people who do?

geekosaur:
This was precipitated by someone reporting that if they used `def` for the keybinding function, keys on the numeric keypad that they had deliberately unbound (for, IIRC, grabbing by some other utility) toggled the docks on/off. I don't think they filed an issue; it was discussed in IRC. Sadly the IRC log doesn't have search functionality so it'll be hard to dig out the discussion.


Looks like it's starting here https://ircbrowse.tomsmeding.com/browse/xmonad?events_page=358 and goes on for a couple of pages; I tracked it down here https://ircbrowse.tomsmeding.com/browse/xmonad?events_page=361.

liskin:
Awesome, thanks. I updated the commit message.

src/XMonad/Main.hs
```diff
@@ -473,8 +473,11 @@ grabKeys = do
     -- build a map from keysyms to lists of keysyms (doing what
     -- XGetKeyboardMapping would do if the X11 package bound it)
     syms <- forM allCodes $ \code -> io (keycodeToKeysym dpy code 0)
-    let keysymMap = M.fromListWith (++) (zip syms [[code] | code <- allCodes])
-        keysymToKeycodes sym = M.findWithDefault [] sym keysymMap
+    let keysymMap' = M.fromListWith (++) (zip syms [[code] | code <- allCodes])
+    -- keycodeToKeysym returns noSymbol for all unbound keycodes, and we don't
+    -- want to grab those whenever someone accidentally used def :: KeySym
```
slotThe: perhaps `uses`?

geekosaur: Sounds better to me, yes.

PR #299, opened 2021-05-23
Remove all instances of deriving Typeable

exorcist365:
### Description

https://github.com/xmonad/xmonad-contrib/issues/548

### Checklist

  - [x] I tested my changes with my setup ~~[xmonad-testing](https://github.com/xmonad/xmonad-testing)~~

### DIFF
--- a/src/XMonad/Core.hs
+++ b/src/XMonad/Core.hs
@@ -154,7 +154,7 @@ newtype ScreenDetail = SD { screenRect :: Rectangle }
 -- instantiated on 'XConf' and 'XState' automatically.
 --
 newtype X a = X (ReaderT XConf (StateT XState IO) a)
-    deriving (Functor, Monad, MonadFail, MonadIO, MonadState XState, MonadReader XConf, Typeable)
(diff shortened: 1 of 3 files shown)


CHANGES.md
```diff
@@ -2,6 +2,8 @@
 
 ## unknown (unknown)
 
+  * Remove all instances of `deriving Typeable`
```
slotThe: It might be worth mentioning that this is because GHC derives this automatically now; we're not actually removing any functionality

liskin: Do we really need a changelog entry here?

liskin: My point being: users don't care that we removed some unnecessary statements from our code. It's just noise. The only possibly relevant user-visible change is that we're no longer compatible with GHC < 8.0, but we likely broke that compat earlier elsewhere.

slotThe: Mh that's a good point.  We should just remove this then

exorcist365:
By README I mean CHANGES.md, my bad.

liskin:
Thanks!

Issue #264, opened 2021-01-15
Next release (v0.17.0)


slotThe:
#### pre-release

* [x] prepare https://github.com/xmonad/xmonad-web/pull/40 to be mergeable (finalize the date, push to liskin.github.io/xmonad-web/, run linkchecker, …)
* [x] packdeps
* [x] version numbers in xmonad.cabal
* [x] version number and date in CHANGES.md
* [x] https://github.com/xmonad/xmonad-contrib/issues/393 pre-release

#### release

* [x] create release in GitHub UI, wait for CI
* [x] check candidate on Hackage and its docs
* [x] manually run the Haskell-CI workflow again to finalise the Hackage release
* [x] https://github.com/xmonad/xmonad-contrib/issues/393 release
* [x] merge https://github.com/xmonad/xmonad-web/pull/40
* [x] drop/change the "we're working hard on the 0.17 release" at https://github.com/sponsors/xmonad
* [x] announcement to `{xmonad,haskell-cafe}@haskell.org`
* [x] announcement to https://discourse.haskell.org/
* [x] announcement to Twitter
* [x] announcement to https://old.reddit.com/r/xmonad/

#### post-release
* [ ] maybe bump it in Stackage?

### Original issue body

> ### Problem Description
>
> I think that when xmonad/xmonad-contrib/issues/393 hits we should also
prepare another release of xmonad itself, so these two new versions may
be used with each other.
>
> There hasn't been a lot happening here, but in particular it would be
nice to include #258 (though I'm obviously biased there) in a release as
soon as possible.  I'm no expert on whether this should then be a minor
release or anything like that, so I'll leave that to people who actually
know what they are talking about.

liskin:
:+1: 

It's worth noting that https://github.com/xmonad/xmonad/commit/735fb58f6cc2b52fcc2c8b24d5e1d3e127bae86e and https://github.com/xmonad/xmonad/commit/6dc1e319d177ac9aa6b9b6f6395077827ee7e1a3 are breaking changes that need an xmonad-contrib release first. Without these, we could just release the few fixes here separately. You all probably knew this but it's better to have it explicitly stated.

Also, I'd love to gather feedback on #260 from the Debian package maintainers (who were Cc:d on the mailing list thread) and resolve the stalemate here. This, I believe, will need @byorgey's involvement.

[...]

Issue #75, opened 2017-02-08
Maintainer and Release Documentation


pjones:
As previously recommended by @geekosaur, we need a way to keep track of all the accounts and details associated with the XMonad project.  I'm going to build a list of questions that we need to answer.  I think the best place to keep this information would be in a `MAINTAINERS.md` file in this repository.

  * [x] How does a new maintainer get added to Hackage?
  * [x] What are the steps for updating the xmonad.org website?
  * [x] Who owns/runs the xmonad Twitter account?
  * [x] What are the steps for releasing xmonad?
  * [x] OpenPGP keys for the maintainers
  * [x] Contact details for the maintainers, email, github, IRC, Twitter, etc.

@xmonad/core please help me build and answer these questions.  Thank you.

aavogt:
Hackage Maintainer: see http://hackage.haskell.org/package/xmonad/maintain

Website: I believe pushing to xmonad-web repo is all that needs to be done. @davidlazar set it up

Twitter: dons gave me the password

PGP: I don't have one but it seems easy enough to figure out

Release: see https://github.com/xmonad/xmonad/blob/master/TODO . It could be more detailed (`git tag ... ; cabal sdist; cabal upload dist/xmonad-...`; `/msg nickserv ???` `/topic #xmonad ....` )

Contact: vogt . adam @ gmail .com. aavogt/enthropy on freenode but I'm not on much lately

geekosaur:
`/msg chanserv topic #xmonad ....` assuming you are authed with nickserv and have the requisite channel access level. (see `/msg chanserv access #xmonad list`)

pjones:
To close this ticket someone needs to finish this file:

https://github.com/xmonad/xmonad/blob/master/MAINTAINERS.md

It still needs information about where all the parts of xmonad live (web, wiki, etc.), how add someone to hackage, how to add someone to IRC, how to add someone to GitHub, etc.

dmwit:
* Adding someone to Hackage: visit http://hackage.haskell.org/package/xmonad/maintainers/ for a list of people authorized to give permissions to the xmonad package. There is an \[edit\] link there to add more folks. Similarly for the other packages (xmonad-contrib, xmonad-extras, X11, X11-xft).
* Adding someone to IRC: `/msg chanserv access add #xmonad <nick> +ARefiorstv`
[...]

PR #308, opened 2021-06-07
Automate and document releasing to Hackage + couple cleanups

liskin:
#### [Revise dependencies a bit](../commit/96baa79c3548af02c95f90a1b907e50740c8680a)

utf8-string isn't needed since d9e3ebf53194d1cf02bae28f0c885b5c75610cfa (late 2014).

Most of the executable xmonad deps are unnecessary since 307b82a53d519f5c86c009eb1a54044a616e4a5c (early 2015).

Switch from data-default to data-default-class to avoid bringing in deps and instances we don't need.

#### [Revise extra-source-files in xmonad.cabal](../commit/061b1bde977233a5b7ad2656c900f0d981885733)

Ship the new docs, drop tests as these get picked up automatically.

#### [Automate and document releasing to Hackage](../commit/9640ca7077e771d54272f6e0ce923085c81d3c77)

Pioneered in https://github.com/xmonad/X11, this adds automation for Hackage releases and updates MAINTAINERS.md with simplified instruction for the release procedure.

Related: https://github.com/xmonad/xmonad/issues/75
Related: https://github.com/xmonad/xmonad/issues/264
Related: https://github.com/xmonad/xmonad-contrib/issues/393

### Checklist

### DIFF
--- a/MAINTAINERS.md
+++ b/MAINTAINERS.md
@@ -59,58 +59,59 @@
 
 ## Release Procedures
 
-When the time comes to release another version of XMonad and Contrib...
+When the time comes to release another version of xmonad and xmonad-contrib:
 
-  1. Create a release branch (e.g., `release-0.XX`).
+  1. Update the version number in all the `*.cabal` files and let the CI
+     verify that it all builds together.
 
-     This will allow you to separate the release process from main
-     development.  Changes you make on this branch will be merged back
-     into `master` as one of the last steps.
+  2. Review documentation files and make sure they are accurate:
 
-  2. Update the version number in the `*.cabal` files and verify
-     dependencies and documentation.  This includes the `tested-with:`
-     field.
+     - [`README.md`](README.md)
+     - [`CHANGES.md`](CHANGES.md)
+     - [`INSTALL.md`](INSTALL.md)
+     - [`man/xmonad.1.markdown.in`](man/xmonad.1.markdown.in)
+     - [haddocks](https://xmonad.github.io/xmonad-docs/)
 
-  3. Use the [packdeps][] tool to ensure you have the dependency
-     versions correct.  If you need to update the version of a
-     dependency then you should rebuild and retest.
+     If the manpage changes, wait for the CI to rebuild the rendered outputs.
 
-  4. Review documentation files and make sure they are accurate:
+  3. Check dependency bounds. In particular, [packdeps][] can be used to check
+     if any new versions are excluded by the upper bounds specified.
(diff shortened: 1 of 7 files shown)


MAINTAINERS.md
```diff
@@ -59,58 +59,58 @@
...
+  2. Review documentation files and make sure they are accurate:
 
-  2. Update the version number in the `*.cabal` files and verify
-     dependencies and documentation.  This includes the `tested-with:`
-     field.
+     - [`README.md`](README.md)
+     - [`CHANGES.md`](CHANGES.md)
+     - [`INSTALL.md`](INSTALL.md)
+     - [`man/xmonad.1.markdown.in`](man/xmonad.1.markdown.in)
+     - haddocks <!-- TODO: link to web with doc snapshots once there's one -->
```
slotThe: There is one now! :)

MAINTAINERS.md
```diff
@@ -59,58 +59,58 @@
...
-  9. Merge the release branches into `master`
+     - Post a [new release announcement][web-announce]
+     - Check install instructions, guided tour, keybindings cheat sheet, …
 
-  10. Update the website:
+  7. Send e-mail announcement to:
 
-      * Generate and push haddocks with `xmonad-web/gen-docs.sh`
+     - [XMonad mailing list](https://mail.haskell.org/mailman/listinfo/xmonad)
+     - [Haskell Cafe](https://mail.haskell.org/cgi-bin/mailman/listinfo/haskell-cafe)
```
slotThe, line 105: Perhaps another cool place would be the Haskell Discourse

slotThe (approved):
This is awesome!  Thank you for improving the release process to this degree

MAINTAINERS.md
liskin: Yes, I will squash the fixup before merging :-)

MAINTAINERS.md
liskin: daf6d06

MAINTAINERS.md
slotThe: Oh!  Didn't realize there was a fixup already, sorry about that

liskin:
I shall note that the release workflow here is somewhat different from the one that was used for the X11 release because there was an issue that needed to be fixed, and thus what is here is (to some extent) untested. We will have one opportunity to test it when releasing xmonad, and then another with xmonad-contrib. Then I'll make sure all workflows are the same.

I'll merge it now as there is nothing to be done before trying it on the actual release.

PR #309, opened 2021-06-20
Use `printf` instead of `echo` when using `xmessage`

MuhammedZakir:
### Description

When using `echo`, the whole message is shown in a single
line because newline is printed as "\n". We could fix this
by either removing `show` and explicitly adding `\"` before
and after the string, or, using `printf` instead of `echo`. I used
second one as it looks cleaner.

### Checklist

I am not sure if I should add this to `CHANGES.md`. I will add this if you say so.

### DIFF
--- a/man/xmonad.hs
+++ b/man/xmonad.hs
@@ -129,7 +129,7 @@ myKeys conf@(XConfig {XMonad.modMask = modm}) = M.fromList $
     , ((modm              , xK_q     ), spawn "xmonad --recompile; xmonad --restart")
 
     -- Run xmessage with a summary of the default keybindings (useful for beginners)
-    , ((modm .|. shiftMask, xK_slash ), spawn ("echo \"" ++ help ++ "\" | xmessage -file -"))
+    , ((modm .|. shiftMask, xK_slash ), spawn ("printf " ++ show help ++ " | xmessage -file -"))
     ]
     ++
 
--- a/src/XMonad/Config.hs
+++ b/src/XMonad/Config.hs
@@ -239,7 +239,7 @@ keys conf@(XConfig {XMonad.modMask = modMask}) = M.fromList $
         , (f, m) <- [(W.view, 0), (W.shift, shiftMask)]]
   where
     helpCommand :: X ()
-    helpCommand = spawn ("echo " ++ show help ++ " | xmessage -file -")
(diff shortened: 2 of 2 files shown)


slotThe:
> When using `echo`, the whole message is shown in a single
> line because newline is printed as "\n". 

Does it?  Seems to work for me: 

2021-06-20-135812_1361x746_scrot

MuhammedZakir:
I am using Manjaro (KDE Plasma), but only running XMonad.

These don't work:

```
"echo " ++ show help ++ " | xmessage -file -"
"echo " ++ help ++ " | xmessage -file -"
```

These work:
```
"echo \"" ++ help ++ "\" | xmessage -file -"
"printf " ++ show help ++ "xmessage -file -"
```

MuhammedZakir:
From echo (GNU) manpage:
```
-e     enable interpretation of backslash escapes
-E     disable interpretation of backslash escapes (default)
```

When I passed `-e` flag to `echo`, it worked as expected. So for help function, either add `-e` flag, or use `printf`. I still stand for `printf` as it is more portable. See https://unix.stackexchange.com/questions/65803/why-is-printf-better-than-echo/#65819.

Edit: I forgot it actually executes within a shell, so the `echo` used must be a shell builtin, right? `spawn` uses `/bin/sh`, which is symlinked to bash in my system. By default, Bash's echo doesn't interpret backslash escapes (`-e` flag is needed). So that explains why `\n` is shown. I guess yours is symlinked to a different shell? E.g. Zsh interprets them by defaut.

slotThe:
> Edit: I forgot it actually executes within a shell, so the `echo` used must be a shell builtin, right? `spawn` uses `/bin/sh`, which is symlinked to bash in my system. By default, Bash's echo doesn't interpret backslash escapes (`-e` flag is needed). So that explains why `\n` is shown. I guess yours is symlinked to a different shell? E.g. Zsh interprets them by defaut.

I'm using dash as `/bin/sh`, which apparently does not have `echo` as a builtin: 

``` console
$ which echo
/bin/echo
```

I can indeed reproduce this if I set the default shell to bash.  Checking real quick, it seems that NixOS also (unwisely) links `/bin/sh` to bash, so this can indeed be a problem for many users.  I don't have a strong opinion on `echo` vs. `printf`, though the argumentation in that stackexchange post seems convincing.  Thanks!

PR #314, opened 2021-07-26
INSTALL: Document xsession/xinitrc, cabal-install

liskin:
### Description

These bits were missing from INSTALL.md and the existing instructions for cabal don't really work any more, so I tried to fill this in.

I also wanted to clean up README.md, drop stuff that's duplicated elsewhere and link to the relevant pieces instead, but I ran out of steam, so I'm posting this for review as it is. If anyone wants to tackle the README bits, and possibly also any mentions of cabal on the website, that would be most welcome. If not, I'll get to it later, maybe, I guess. :-)

### Checklist

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -18,6 +18,7 @@ Those who install from distro can skip this and go straight to
 - [Download XMonad sources](#download-xmonad-sources)
 - [Build XMonad](#build-xmonad)
   - [Build using Stack](#build-using-stack)
+  - [Build using cabal-install](#build-using-cabal-install)
 - [Make XMonad your window manager](#make-xmonad-your-window-manager)
 - [Don't Recompile on Every Startup](#dont-recompile-on-every-startup)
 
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -229,6 +230,56 @@ And that's it!  Recompilation should work normally now, though you will
...
 session, first.
 
+### Build using cabal-install
+
+#### Install cabal-install
+
+The easiest way to get [cabal-install] is probably via your system's package
+manager:
+
+``` shell
```
slotThe: I wonder if we could get nicer syntax highlighting via `console` here (and in other places) or if that doesn't play nicely with the comments

INSTALL.md
```diff
@@ -1,24 +1,47 @@
 # Install XMonad
 
+On many systems xmonad is available as a binary package in your
+distribution (Debian, Ubuntu, Fedora, Arch, Gentoo, …).
+It's by far the easiest way to get xmonad, although you'll miss on the latest
```
slotThe:
``` suggestion
It's by far the easiest way to get xmonad, although you'll miss out on the latest
```

slotThe (approved):

INSTALL.md
liskin: It definitely does make the `$ stack init` block easier on the eyes: https://github.com/xmonad/xmonad/blob/62069e8af70de7793e7541eaf4c6381521469456/INSTALL.md#create-a-new-project vs https://github.com/xmonad/xmonad/blob/11e41b98d151662627efba16cc4d3eb00ae2d8b6%5E%5E/INSTALL.md#create-a-new-project
Most other blocks look the same. I had to add leading `> ` to [multiline commands in dependencies](https://github.com/xmonad/xmonad/blob/62069e8af70de7793e7541eaf4c6381521469456/INSTALL.md#dependencies) because the `console` highlighter would otherwise [highlight the continuation lines as output](https://github.com/xmonad/xmonad/blob/6b3a0048f7719124c8b42772ef7d3acded99ac03/INSTALL.md#dependencies).

liskin: Overall it's probably better so I'll wait a couple hours for you to see it and then rebase and merge. Pls ping me on IRC or here then.

PR #316, opened 2021-07-31
Add a flake.nix file

### LINKED ISSUE #315: Add a flake.nix file so that xmonad can be easily used as a flake in nix-builds
Is this something the maintainers would be open to? I have one ready to go.

colonelpanic8:
### Description

behind them.

### Checklist

Fixes #315

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -1,0 +1,25 @@
+name: Nix
+
+on:
+  push:
+  pull_request:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Nix Flake - Linux
+    permissions:
+      contents: read
+    steps:
+      - name: Install Nix
+        uses: cachix/install-nix-action@v13
+        with:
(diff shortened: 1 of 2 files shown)


colonelpanic8:
BTW, guarantee that the build we could do to test this using nix flakes would be like 10x faster than the current build, since we would get all deps from nix-cache.

.github/workflows/nix.yml
```diff
@@ -21,3 +23,5 @@ jobs:
         uses: actions/checkout@v2
       - name: Build
         run: nix build
+      - name: Check
+        run: nix flake check
```
colonelpanic8: This wont work because of a bug in how IFD is handled in these properties right now.

see https://github.com/NixOS/nix/issues/4265

liskin: Oh, you were quicker than me, I was just writing a comment asking you whether that's expected or not. I'll drop it for now then, looks like you do know what you're doing. Good! :-)

colonelpanic8: Yeah I would have added the `nix check` myself if not for this issue.

I think there are plans to allow nix check of specific fields, so we can at least do that if the main issue is not fixed when that happens.

Issue #178, opened 2018-08-05
Should the generated files really go into the config directory?


xaverdh:
### Problem Description

At the moment generated files (hi, object files) go into the config directory.
This is true even if all XDG_*_HOME variables are set.
Now in some cases it seems desirable to store the source files in a separate location from the arch specific files. My use case would be sharing the configuration between several machines by just mirroring the directory.

As a side note, although I am not too familiar with the xdg spec, I wonder if putting the hi/object files into the config directory is consistent with the specification.

liskin:
This is a good point actually, we might want to add `-outputdir ~/.local/share/xmonad` to the `ghc` command-line.

liskin:
Implemented in https://github.com/xmonad/xmonad/pull/318 if you'd like to help with testing/review :-)

PR #318, opened 2021-08-07
Recompilation overhaul: stack.yaml detection, deprecation warnings

### LINKED ISSUE #304: Deprecation warnings are suppressed during recompilation
### Problem Description

Deprecation warnings are suppressed during recompilation, which makes it difficult to improve things in the codebase and API as users do not (know they) have a transition period to adapt their configuration.

See https://github.com/xmonad/xmonad-contrib/pull/404#issuecomment-731607447 and https://github.com/xmonad/xmonad-contrib/pull/410#issuecomment-732841235 for some ideas how to deal with this.

### LINKED ISSUE #310: Native recompilation support for stack/cabal installed XMonad
A cli option for this would be nice. Simple support is enough. That is, for example, `xmonad --tool stack`, will run `stack ghc ...` instead of `ghc ...`. If a user needs to compile in a different way (e.g. with different flags), they can always use build script.

### Checklist

  - [ ] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

liskin:
### Description

Addresses https://github.com/xmonad/xmonad/issues/304 and https://github.com/xmonad/xmonad/issues/310 and https://github.com/xmonad/xmonad/issues/178.

Probably needs a thorough review and possibly some more testing, as I definitely didn't test this very extensively.

And the deprecation warnings detection is probably a bit less robust than I'd like. The more robust alternative is explicitly invoking GHC with `-Werror=deprecations`, but I couldn't decide when to invoke it. :-)

#### [Refactor 'recompile' to be less of a spaghetti-code](../commit/90240a23df9906d660c604ce74f29ae90bf408d2)

This is a preparation for autodetecting stack.yaml and using stack for
recompilation.

Related: https://github.com/xmonad/xmonad/issues/310

#### [Make xmessage handle UTF-8 and export it](../commit/9f54afe793ddb02923b0f4fb468c26c0aab7ead6)

This unfortunately breaks xmonad-contrib as several modules define their own `xmessage` function.

Related: https://github.com/xmonad/xmonad/pull/309

#### [Use "stack ghc" for recompilation when stack.yaml exists](../commit/e8f40cf8966ce4a0d5bef3ad3d287129f5a431bd)

This makes it unnecessary for users of Stack to use a custom build script, making installation easier.

Inspired by a similar feature in dyre: https://github.com/willdonnelly/dyre/commit/a04be85f60cda39b3d57cebb74a25e9508b58160

Fixes: https://github.com/xmonad/xmonad/issues/310

#### [Move config binary and GHC intermediate outputs to cacheDir](../commit/7aad1ff3cbd74c0d1f177502454ff3d7ee1fb0dc)

That's where they belong. As XDG was more or less broken in previous xmonad releases, we can assume few people use it so now's the best time to move files around.

For users of `~/.xmonad`, this only causes intermediate outputs (.o, .hi) to go elsewhere.

#### [Fix indent in getDirectories](../commit/ac5c096880e18fb9f0a801830c5cfc20533d8ec9)

Fixes: 90101613e76d ("Unclobber dirs/Dirs in import XMonad")

#### [Detect deprecation warnings during recompilation and warn the user](../commit/97ae1489bdc3252f7024bb52993812474f442f90)

Deprecation warnings were suppressed during recompilation, which made it difficult to improve things in the codebase and API as users did not (know they) have a transition period to adapt their configuration.

See additional discussions about deprecations:

 - https://github.com/xmonad/xmonad-contrib/pull/404#issuecomment-731607447
 - https://github.com/xmonad/xmonad-contrib/pull/410#issuecomment-732841235

Fixes: https://github.com/xmonad/xmonad/issues/304
Related: https://github.com/xmonad/xmonad-contrib/pull/404
Related: https://github.com/xmonad/xmonad-contrib/pull/410

#### [INSTALL: Update after stack.yaml autodetection](../commit/eb04c21940d108e2760e0059eb9d570bd029e594)

Related: https://github.com/xmonad/xmonad/issues/310

### Checklist


src/XMonad/Core.hs
```diff
@@ -552,19 +567,149 @@ getXMonadDataDir :: X String
...
+            trace $ "XMonad will use build script at " <> show buildScript <> " to recompile."
+            pure $ CompileScript buildScript
+          else do
+            trace $ "XMonad will not use build script, because " <> show buildScript <> " is not executable."
+            trace $ "Suggested resolution to use it: chmod u+x " <> show buildScript
+            empty
+
+    tryStack = do
+        guard =<< doesFileExist stackYaml
+        canonStackYaml <- canonicalizePath stackYaml
```
TheMC47, line 612: Isn't this a bit of an overkill? Won't a simple `makeAbsolute` suffice?

TheMC47 (commented):
Thank you for the much-needed overhaul and improvements :tada: The code looks way less intimidating now, to the point where I could decipher exactly what's going on.

This looks good, I haven't tested it yet (I'll probably get to it in the next couple of days)

src/XMonad/Core.hs
liskin: I specifically wanted this to resolve symlinks so that people can place the `stack.yaml` file and the git clones elsewhere. Having sources and `.stack-work`s inside of `~/.config` feels wrong and some people won't like it.

TheMC47: Oh alright then, that makes sense

src/XMonad/Core.hs
```diff
@@ -552,19 +567,149 @@ getXMonadDataDir :: X String
...
+                <> show buildScript <> " nor " <> show stackYaml <> " exists."
+        pure CompileGhc
+
+    isExecutable f = E.catch (executable <$> getPermissions f) (\(SomeException _) -> return False)
+
+-- | Should we recompile xmonad configuration? Is it newer than the compiled
+-- binary?
+shouldCompile :: Directories -> Compile -> IO Bool
+shouldCompile dirs CompileGhc = do
+    libTs <- mapM getModTime . Prelude.filter isSource =<< allFiles (libFileName dirs)
```
xaverdh, line 627: This is probably a bit out of scope of this pr, but remembering a hash of the sources (vs comparing mod. time) is the right thing to do here I would say.

src/XMonad/Core.hs
```diff
@@ -453,6 +454,17 @@ xfork x = io . forkProcess . finally nullStdin $ do
...
         closeFd fd
 
+-- | Use @xmessage@ to show information to the user.
+xmessage :: MonadIO m => String -> m ()
+xmessage msg = void . xfork $ do
+    setEnv "LC_ALL" "C.UTF-8"
+    executeFile "xmessage" True
+        [ "-default", "okay"
+        , "-xrm", "*international:true"
+        , "-xrm", "*fontSet:-*-fixed-medium-r-normal-*-18-*-*-*-*-*-*-*,-*-fixed-*-*-*-*-18-*-*-*-*-*-*-*,-*-*-*-*-*-*-18-*-*-*-*-*-*-*"
```
slotThe, line 466: I wonder if we'll run into https://github.com/xmonad/xmonad-contrib/issues/348 at some point when distributions start dropping PCF fonts altogether

src/XMonad/Core.hs
```diff
@@ -670,6 +671,22 @@ compile dirs method =
...
+checkCompileWarnings dirs = do
+    ghcErr <- readFile (errFileName dirs)
+    if "-Wdeprecations" `isInfixOf` ghcErr
+      then do
+        let msg = unlines $
+                ["Deprecations detected while compiling xmonad config: " <> srcFileName dirs]
+                ++ lines ghcErr
+                ++ ["","Please correct them or silence using {-# OPTIONS_GHC -Wno-deprecations #-}."]
+        trace msg
+        xmessage msg
```
slotThe, line 701: I wonder if we should do something a little bit more clever here and extract the actual deprecation warnings in order to not overwhelm the user with the whole compilation output

slotThe (commented):
Based on a cursory first look this is very nice :+1:

My personal configuration "just worked" with the stack recompilation, after deleting the build script and moving `xmonad.hs` to the right location; will test this in other situations in the next few days

src/XMonad/Core.hs
liskin: At that point `xmessage` stops working entirely, even without these options. We'll need to check for zenity/gxmessage/whatever using [findExecutable](https://hackage.haskell.org/package/directory-1.3.6.2/docs/System-Directory.html#v:findExecutable) then.

src/XMonad/Core.hs
liskin: It's invoked with `-v0` and without `-Wall`, so there isn't a whole lot of “compilation output” to speak about. :-)

src/XMonad/Core.hs
liskin: It'd be more precise but probably not worth the complexity. I know `stack` and `nix` likely do it, and perhaps `shake` has support for it, but `make` worked for decades without it. :-)

src/XMonad/Core.hs
slotThe: Ah good point; especially since we're only compiling the file itself

src/XMonad/Core.hs
```diff
@@ -552,19 +567,149 @@ getXMonadDataDir :: X String
...
+    isSource = flip elem [".hs",".lhs",".hsc"] . takeExtension
+    allFiles t = do
+        let prep = map (t</>) . Prelude.filter (`notElem` [".",".."])
+        cs <- prep <$> E.catch (getDirectoryContents t) (\(SomeException _) -> return [])
+        ds <- filterM doesDirectoryExist cs
+        concat . ((cs \\ ds):) <$> mapM allFiles ds
+shouldCompile dirs CompileStackGhc{} = do
+    stackYamlT <- getModTime (stackYamlFileName dirs)
+    binT <- getModTime (binFileName dirs)
+    if binT < stackYamlT
```
slotThe, line 643: I wonder if we should perhaps also execute `stack build` here?  If something in the stack yaml changes, it's likely that this is either adding a new dependency, or bumping the resolver.  In this case, we should rebuild everything before compiling the config, as we would get errors otherwise (because e.g. `xmonad-contrib` is just not built for the new resolver) and would have to manually invoke `stack build` (or `stack install` or something similar).  This might not be so obvious behaviour to users

liskin: Oh, indeed! I don't know why I thought `stack ghc` would do this automatically. Probably just tired. :-)

liskin: 3f6569db66f3a2ac51005258a8ba1544d5ae1a6a should do it

liskin:
@slotThe Did you get the following resolved?

```
12:43:16 <Solid> the xmonad M-q invocation didn't work either :/
```

Are there any other pending issues here? Do we need some more docs about the stack.yaml symlink perhaps?

src/XMonad/Core.hs
slotThe: I'm now getting a bunch of 

    xmonad: stack: runProcess: invalid argument (Bad file descriptor)

which is especially weird since both 

    stack build --silent

and 

    stack ghc --stack-yaml stack.yaml -- --make xmonad.hs -i -ilib -fforce-recomp -main-is main -v0

work just fine when entered from the console.  I'll try to debug this later today

slotThe: Ah I think the reason is that [runProcess](https://hackage.haskell.org/package/process-1.6.13.2/docs/src/System.Process.html#runProcess) closes the handles that it gets: 

> Any Handles passed to runProcess are placed immediately in the closed state.

and so the second invocation fails because it expects `xmonad.errors` to be open

liskin: Oh, sorry, I didn't bother running the code, and of course it bit me. We need linear types! :-)

> Any Handles passed to runProcess are placed immediately in the closed state.

I'll fix this.

liskin: 79682c5bb144ba7b872a07823feb5618e52d301b works for me

slotThe: Works :+1:  

The error messages this generates when something goes bad are quite funny 

``` console
Errors detected while compiling xmonad config: /home/slot/.config/xmonad/xmonad.hs
ExitFailure 1

Please check the file for errors.
```

but I'm not sure we can do anything about that (this is just a byproduct of the `--silent` flag).  

All in all this looks good to me now (and I also can't reproduce the problem mentioned below anymore)

liskin: It might be quite helpful to print the executed commands into the error file to make its output less cryptic. I'll add that and merge this now to get as much testing as possible.

PR #326, opened 2021-09-02
Make extensibleState primarily keyed by TypeRep instead of type names

### LINKED ISSUE #94: X.U.ExtensibleState and typename as map key
It's useless when compared with using just any string as map key because two different modules could have wrapper with the same name, and checking for that is not easier then checking for same string as key. It does also require these annoying wrappers and unwrappers for no benefit. Should we migrate to using any string as keys or is it useless or even wrong?

liskin:
### Description

We've been using the String we get out of `show . typeOf` as key in `extensibleState`, but that has a somewhat serious bug: it shows unqualified type names, so if two modules use the same type name, their extensible states will be stored in one place and get overwritten all the time.

To fix this, the `extensibleState` map is now primarily keyed by the TypeRep themselves, with fallback to String for not yet deserialized data. XMonad.Core now exports `showExtType` which serializes type names qualified, and this is used in `writeStateToFile`.

A simpler fix would be to just change the serialization of type names in `XMonad.Util.ExtensibleState`, but I'm afraid that might slows things down: Most types used here will start with "XMonad.", and that's a lot of useless linked-list pointer jumping.

Fixes: https://github.com/xmonad/xmonad-contrib/issues/94
Related: https://github.com/xmonad/xmonad-contrib/pull/600

---

I must admit I don't really like the code, especially the xmonad-contrib part. There's the assumption that a Right key maps to a Right value, and Left key to Left value, but enforcing that in the types probably means adding a another field to the XState.

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -65,6 +65,13 @@
     it easier for us to clean up the codebase. These can still be suppressed
     manually using an `OPTIONS_GHC` pragma with `-Wno-deprecations`.
 
+  * Extensible state type names no longer need to be unique, because the
+    `extensibleState` map in `XState` is now primarily keyed by the
+    machine-readable type representation rather than the human-readable type
+    name. Human-readable type names are still used for serialization of state
+    between restarts, and this representation now encodes module names as
+    well to avoid conflicts between types with equal names.
+
 ## 0.15 (September 30, 2018)
 
   * Reimplement `sendMessage` to deal properly with windowset changes made
--- a/src/XMonad/Core.hs
+++ b/src/XMonad/Core.hs
@@ -1,6 +1,13 @@
-{-# LANGUAGE ExistentialQuantification, FlexibleInstances, GeneralizedNewtypeDeriving,
-             MultiParamTypeClasses, TypeSynonymInstances, DeriveDataTypeable,
-             LambdaCase, NamedFieldPuns, DeriveTraversable #-}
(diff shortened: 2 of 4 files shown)


slotThe:
> A simpler fix would be to just change the serialization of type names in XMonad.Util.ExtensibleState, but I'm afraid that might slows things down: Most types used here will start with "XMonad.", and that's a lot of useless linked-list pointer jumping.

As a great man once told me: have you benchmarked this?  We still use `String` all over the place, so I can't imagine that a linked list of length 7 has that much of a performance impact

liskin:
I didn't, true. It just feels wrong, and a couple other people suggested using [Fingerprint](https://hackage.haskell.org/package/base-4.15.0.0/docs/GHC-Fingerprint.html#t:Fingerprint) in the past, so I went ahead and tried implementing it. It may indeed be negligible in the grand scheme of things (although that would suggest other things are also terribly inefficient). I'll try to benchmark this somehow in the coming days.

geekosaur:
Turns out Fingerprint (and therefore the internal representation of TypeRep) is not stable across compiles, much less GHC versions, so we would need to use the String anyway unless you deal with this during serialization and deserialization. I'd just go ahead and use the String, possibly with "XMonad." prefix removed (although this is a problem if we're passed a type that isn't defined by xmonad).

geekosaur (approved):
LGTM

PR #330, opened 2021-09-11
Add nix recompilation support

colonelpanic8:
### Description

This is a follow up to #318 that adds support for nix recompilation out of the box

### Checklist

I tested flake recompilation with my config and it seems to work well.

### DIFF
--- a/src/XMonad/Core.hs
+++ b/src/XMonad/Core.hs
@@ -55,6 +55,7 @@ import Data.Traversable (for)
 import Data.Time.Clock (UTCTime)
 import Data.Default.Class
 import System.Environment (lookupEnv)
+import Data.List (isInfixOf, intercalate, (\\))
 import System.FilePath
 import System.IO
 import System.Info
@@ -69,7 +70,6 @@ import System.Exit
 import Graphics.X11.Xlib
 import Graphics.X11.Xlib.Extras (getWindowAttributes, WindowAttributes, Event)
 import Data.Typeable
-import Data.List (isInfixOf, (\\))
 import Data.Maybe (isJust,fromMaybe)
 import Data.Monoid (Ap(..))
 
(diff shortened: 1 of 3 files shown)


colonelpanic8:
@liskin any idea why the build suddenly errored. Seems unrelated to my change.

liskin:
> @liskin any idea why the build suddenly errored. Seems unrelated to my change.

You've made the `bracket` import redundant.

colonelpanic8:
> > @liskin any idea why the build suddenly errored. Seems unrelated to my change.
> 
> You've made the `bracket` import redundant.

Right I'm just not sure how that happened.

colonelpanic8:
Ah I see. I just applied all the language server hints and one of them seems to have rendered that import redundant.

slotThe:
We should probably also add a `Build using Nix` style section to [INSTALL.md](https://github.com/xmonad/xmonad/blob/master/INSTALL.md#build-xmonad)

src/XMonad/Core.hs
```diff
@@ -607,18 +616,37 @@ detectCompile dirs = tryScript <|> tryStack <|> useGhc
...
+      guard =<< doesFileExist flakeNix
+      canonNixFlake <- canonicalizePath flakeNix
+      trace $ "XMonad will use nix flake at " <> show canonNixFlake <> " to recompile"
+      pure $ CompileNixFlake canonNixFlake
+
+    tryNixDefault = do
+      guard =<< doesFileExist defaultNix
+      canonNixFlake <- canonicalizePath defaultNix
+      trace $ "XMonad will use nix flake at " <> show canonNixFlake <> " to recompile"
+      pure $ CompileNixDefault canonNixFlake
```
MuhammedZakir: `canonNixFlake` should be `canonNixDefault` or `cananDefaultNix`?

psibi:
> We should probably also add a Build using Nix style section to INSTALL.md

Yeah, I agree. I'm using NixOS and I'm using a different workflow and I cannot understand the motivation of these changes. Is this support meant for nix in non NixOS distributions ?

colonelpanic8:
> Yeah, I agree. I'm using NixOS and I'm using a different workflow and I cannot understand the motivation of these changes. Is this support meant for nix in non NixOS distributions?

In #316 I added a nix flake for the xmonad repo. This made it so that there is an easy way to use xmonad master (instead of using the version in hackage-packages.nix which is just the latest version from hackage, which is quite old at this point). That change can work with using the xmonad MODULE provided by nixpkgs if you simply use an overlay to override the xmonad library, but its also totally compatible with other methods of using xmonad together with nix.

I do not like the way that the nixpkgs Xmonad module works, so I use a different approach where I simply have a normal haskell project with some minimal nix infrastructure around it, that allows the building of my xmonad executable. I'm not alone in using this approach (many actually do something similar). As you seem to have guessed, this approach also has the advantage that it works outside of nixos.

This change arose because @liskin was upset that we have an official flake for the repository, but the flake is not necessarily compatible with self recompliation, which he recently overhauled to work with stack in  #318 (see https://github.com/alternateved/nixos-config/issues/1#issuecomment-917248073).

psibi:
Ah, I see. Thanks for the background information.

> This made it so that there is an easy way to use xmonad master (instead of using the version in hackage-packages.nix which is just the latest version from hackage, which is quite old at this point).

Note that I use the latest stuff in git using nix overrides. I haven't yet started using flakes as last time I checked it was still marked experimental. But yeah, it's nice to have support for it. :-)

> I do not like the way that the nixpkgs Xmonad module, works, so I use a different approach where I simply have a normal haskell project with some minimal nix infrastructure around it, that allows the building of my xmonad executable.

Can we probably have this documented under the install section ? I have no idea how this works and can review the PR too. 

> As you seem to have guessed, this approach also has the advantage that it works outside of nixos.

Note that I did use the overrides approach before in my Ubuntu machine before I completely moved on to NixOS.

colonelpanic8:
> Can we probably have this documented under the install section ? I have no idea how this works and can review the PR too.

I'm not sure its necessary to document the peculiarities of how I personally use this. The point is we have a nix flake that works just like any nix flake works. The nix flake documentation should suffice.



> I haven't yet started using flakes as last time I checked it was still marked experimental. But yeah, it's nice to have support for it. :-)

It's not that experimental in practice at this point since its used so widely. I think major changes to the way flakes are defined and used are somewhat unlikely at this point.

Issue #208, opened 2020-01-09
Notification windows not being placed on top of floating windows


tsipinakis:
Disclaimer: I'm not an xmonad user, this is the a forward of dunst-project/dunst#678, accurate steps to reproduce the issue are included in that issue, it works with the default config.

This issue seems similar, but not quite the same as #89.

### Problem Description

See dunst-project/dunst#678.

On the dunst side we call `XRaiseWindow` on `CreateNotify` events of non-override-redirect windows. I tried to make that a bit more aggressive so it raises on `VisibilityNotify`'s as well but all that does is it makes dunst and xmonad fight over who'll be on top.

### Configuration File

```haskell
module Main (main) where

import XMonad

main :: IO ()
main = xmonad def
```

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [ ] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

geekosaur:
Out of the box, xmonad only supports ICCCM so such windows must use ICCCM `override_redirect`. dunst is apparently demanding EWMH support.

Additionally, you'd need some extra configuration with `EwmhDesktops` because currently it doesn't support raising such windows by itself. This is arguably a bug and there may already be a bug open for it. The support is there but must be manually configured to `doIgnore` windows with the EWMH notification window property.

tsipinakis:
> Out of the box, xmonad only supports ICCCM so such windows must use ICCCM override_redirect. dunst is apparently demanding EWMH support.

We [do set](https://github.com/dunst-project/dunst/blob/master/src/x11/x.c#L641-L666) the `override_redirect` attribute and use the `CWOverrideRedirect` mask, is there something I'm missing here?

geekosaur:
Interesting. Getting into a fight with xmonad over the window implies that somehow it is getting managed, but `override_redirect` should prevent that. Does a `manageHook` including

`isInProperty "_NET_WM_WINDOW_TYPE" "_NET_WM_WINDOW_TYPE_NOTIFICATION" --> doIgnore`

prevent it? (I am running xmonad under Mate and am poorly positioned to set up test environments currently.)

Note that above code requires `XMonad.Hooks.ManageHelpers` be imported from xmonad-contrib.

tsipinakis:
I don't know any haskell so I'm unsure on how to write that in the config.

[...]

PR #340, opened 2021-10-22
README: Long overdue update

liskin:
### Description

* add logo and badges to make it prettier (note that I've tweaked the badges a bit compared to what was merged to xmonad-contrib, I'll update it there later)
* revamp installation sections
* update xmonad-contrib reference and move upwards
* drop other useful programs, this is in download.html
* add new Contributing section

Rendered: https://github.com/liskin/xmonad/tree/readme#readme

Also:

* LICENSE: Fix wording to make it detectable by GitHub
* Update .mailmap with even more info from xmonad-contrib

TODO:

* [ ] CONTRIBUTING.md is outdated:
  * [ ] xmonad-testing is a bit desolate so we can't ask people to use it
  * [ ] squashing all commits into one? I'm actually glad people don't read CONTRIBUTING.md otherwise I'd snap because of this
  * [ ] perhaps a short paragraph that people who show they're capable of using git, Haskell and communicating with others will get write to access xmonad-contrib and we'll happily let them maintain their modules and more
* [ ] what about STYLE, CONFIG?
* [ ] xmonad-testing in .github/PULL_REQUEST_TEMPLATE.md

### Checklist

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 # Change Log / Release Notes
 
-## unknown (unknown)
+## 0.17.0 (October ??, 2021)
 
   * Fixed a bug when using multiple screens with different dimensions,
     causing some floating windows to be smaller/larger than the size they
--- a/.mailmap
+++ b/.mailmap
@@ -2,14 +2,17 @@ Adam Plaice <plaice.adam+github@gmail.com>
 Brandon S Allbery KF8NH <allbery.b@gmail.com>
 Brent Yorgey <byorgey@gmail.com> <byorgey@cis.upenn.edu>
 Conrad Irwin <conrad.irwin@gmail.com>
+Daniel Neri <daniel.neri@sigicom.com> <daniel.neri@sigicom.se>
 Daniel Schoepe <daniel.schoepe@gmail.com> <asgaroth_@gmx.de>
 Daniel Wagner <me@dmwit.com> <daniel@wagner-home.com>
 David Glasser <glasser@mit.edu>
 Deven Lahoti <deven.lahoti@gmail.com>
+Devin Mullins <devin.mullins@gmail.com> <me@twifkak.com>
 Don Stewart <dons00@gmail.com> <dons@cse.unsw.edu.au>
 Don Stewart <dons00@gmail.com> <dons@galois.com>
+Felix Springer <felixspringer149@gmail.com> <39434424+jumper149@users.noreply.github.com>
 Gwern Branwen <gwern@gwern.net> <gwern0@gmail.com>
-Lukas Mai <l.mai@web.de> <l.mai@web.de>
+Lukas Mai <l.mai@web.de>
 Marshall Lochbaum <mwlochbaum@gmail.com>
 Michael G. Sloan <mgsloan@gmail.com>
 Neil Mitchell <ndmitchell@gmail.com> <http://www.cs.york.ac.uk/~ndm/>
@@ -23,7 +26,13 @@ Robert Marlow <bobstopper@bobturf.org>
 Robert Marlow <bobstopper@bobturf.org> <robreim@bobturf.org>
 Sam Hughes <hughes@rpi.edu>
 Shae Erisson <shae@ScannedInAvian.com>
+Sibi Prabakaran <sibi@psibi.in>
+Sibi Prabakaran <sibi@psibi.in> <psibi2000@gmail.com>
 Spencer Janssen <spencerjanssen@gmail.com> <sjanssen@cse.unl.edu>
 Timothy Hobbs <tim.thelion@gmail.com>
+Tomas Janousek <tomi@nomi.cz>
+Valery V. Vorotyntsev <valery.vv@gmail.com>
 Vanessa McHale <vamchale@gmail.com> <vanessa.mchale@reconfigure.io>
 Wirt Wolff <wirtwolff@gmail.com>
+
+slotThe <soliditsallgood@mailbox.org> <50166980+slotThe@users.noreply.github.com>
--- a/LICENSE
+++ b/LICENSE
@@ -1,31 +1,28 @@
 Copyright (c) 2007,2008 Spencer Janssen
 Copyright (c) 2007,2008 Don Stewart
+Copyright (c) The Xmonad Community. All rights reserved.
 
-All rights reserved.
+Redistribution and use in source and binary forms, with or without modification,
+are permitted provided that the following conditions are met:
 
-Redistribution and use in source and binary forms, with or without
-modification, are permitted provided that the following conditions
-are met:
+1. Redistributions of source code must retain the above copyright notice,
+this list of conditions and the following disclaimer.
 
-1. Redistributions of source code must retain the above copyright
-   notice, this list of conditions and the following disclaimer.
+2. Redistributions in binary form must reproduce the above copyright notice,
+this list of conditions and the following disclaimer in the documentation
+and/or other materials provided with the distribution.
 
-2. Redistributions in binary form must reproduce the above copyright
-   notice, this list of conditions and the following disclaimer in the
-   documentation and/or other materials provided with the distribution.
+3. Neither the name of the copyright holder nor the names of its contributors
+may be used to endorse or promote products derived from this software without
+specific prior written permission.
 
-3. Neither the name of the author nor the names of his contributors
-   may be used to endorse or promote products derived from this software
-   without specific prior written permission.
-
-THIS SOFTWARE IS PROVIDED BY THE AUTHORS ``AS IS'' AND ANY EXPRESS OR
-IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
-WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
-DISCLAIMED.  IN NO EVENT SHALL THE AUTHORS OR CONTRIBUTORS BE LIABLE FOR
-ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
-DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS
-OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION)
-HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT,
-STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
-ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
-POSSIBILITY OF SUCH DAMAGE.
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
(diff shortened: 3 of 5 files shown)


slotThe (approved):
LGTM!

>       * [ ]  xmonad-testing is a bit desolate so we can't ask people to use it

Agreed.

>       * [ ]  squashing all commits into one? I'm actually glad people don't read CONTRIBUTING.md otherwise I'd snap because of this

:)

>       * [ ]  perhaps a short paragraph that people who show they're capable of using git, Haskell and communicating with others will get write to access xmonad-contrib and we'll happily let them maintain their modules and more

I don't think the current bar to getting access to the repo is all that high[^1], but explicitly mentioning this sounds like a good idea :+1:

One more thing that I would change is to more prominently link to IRC and matrix (I know that community.html already does this, but it couldn't hurt to have this here as well).  Something like

``` diff
diff --git a/CONTRIBUTING.md b/CONTRIBUTING.md
index 05ba5ba..dd5b501 100644
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -34,9 +34,10 @@ Awesome!  Here are a few things to keep in mind:
     nontrivial changes to xmonad.  There are a couple of ways you can
     chat with us:

-    - Post a message to the [mailing list][ml].
+    - Join the [`#xmonad` IRC channel] on `irc.libera.chat` or the
+      official [matrix channel], which is linked to IRC.

-    - Join the `#xmonad` IRC channel on `irc.libera.chat`.
+    - Post a message to the [mailing list][ml].

   * [XMonad.Doc.Developing][xmonad-doc-developing] is a great
     resource to get an overview of xmonad.  Make sure to also check
@@ -155,3 +156,5 @@ each pull request contains just one commit.
 [x11]: https://github.com/xmonad/X11
 [ml]: https://mail.haskell.org/cgi-bin/mailman/listinfo/xmonad
 [xmonad-doc-developing]: https://xmonad.github.io/xmonad-docs/xmonad-contrib/XMonad-Doc-Developing.html
+[`#xmonad` IRC channel]: https://web.libera.chat/#xmonad
+[matrix channel]: https://matrix.to/#/#xmonad:matrix.org
```

>     * [ ]  what about STYLE, CONFIG?

CONFIG is, imo, fully replaced by TUTORIAL.md.  STYLE still seems to be pretty accurate for the core at least, but should probably live inside CONTRIBUTING.md instead.  One thing I have a problem with is

> * New modules should identify the author, and be submitted under the same license as xmonad __(BSD3 license or freer)__.

Emphasis mine.  As I see it, the GPL-3 is "freer" than BSD3, but that's probably not what the original authors had in mind :)  I haven't seen any module that was not licensed under the 3-Clause BSD license, so I'd just drop this.  The same sentence is also found in X.D.Extending.

>     * [ ]  xmonad-testing in .github/PULL_REQUEST_TEMPLATE.md

Just drop this, as we did for contrib.

[^1]: I've been thinking a bit about whether that's a good thing or not.  Commit bits can be abused (as certain recent events have shown us), and with like 20 people having access to contrib already, are we just maximising our attack surface, waiting for something bad to happen?

CHANGES.md
```diff
@@ -1,6 +1,6 @@
 # Change Log / Release Notes
 
-## unknown (unknown)
+## 0.17.0 (October ??, 2021)
```
slotThe, line 3: Actually breaking this down into three parts as your todo suggests might be a good idea!  I suppose what constitutes a breaking change and what's a bug fix can be bikeshedded for eternity, but here's what uncaffinated me came up with.

<details>
  <summary>Following contrib's two-part structure</summary>

  ### Breaking Changes

  * Made `(<&&>)` and `(<||>)` non-strict in their right operand; i.e.,
    these operators now implement short-circuit evaluation so the right
    operand is evaluated only if the left operand does not suffice to
    determine the result.

  * Change `ScreenDetail` to a newtype and make `RationalRect` strict in
    its contents.

  * Added the `extensibleConf` field to `XConfig` which makes it easier for
    contrib modules to have composable configuration (custom hooks, …).

  * `util/GenerateManpage.hs` is no longer distributed in the tarball.
    Instead, the manpage source is regenerated and manpage rebuilt
    automatically in CI.

  * `DestroyWindowEvent` is now broadcasted to layouts to let them know
    window-specific resources can be discarded.

### Bug Fixes and Minor Changes

  * Migrated `X.L.LayoutCombinators.(|||)` into `XMonad.Layout`,
    providing the ability to directly jump to a layout with the
    `JumpToLayout` message.

  * Recompilation now detects `stack.yaml` (can be a symlink) alongside
    `xmonad.hs` and switches to using `stack ghc`. We also updated INSTALL.md
    with instructions for cabal-install that lead to correct recompilation.

    Deprecation warnings during recompilation are no longer suppressed to make
    it easier for us to clean up the codebase. These can still be suppressed
    manually using an `OPTIONS_GHC` pragma with `-Wno-deprecations`.

  * Improve handling of XDG directories.

      1. If all three of xmonad's environment variables (`XMONAD_DATA_DIR,`
         `XMONAD_CONFIG_DIR`, and `XMONAD_CACHE_DIR`) are set, use them.
      2. If there is a build script called `build` (see
         [here](https://github.com/xmonad/xmonad-testing/tree/master/build-scripts)
         for usage examples) or configuration `xmonad.hs` in `~/.xmonad`, set
         all three directories to `~/.xmonad`.
      3. Otherwise, use the `xmonad` directory in `XDG_DATA_HOME`,
         `XDG_CONFIG_HOME`, and `XDG_CACHE_HOME` (or their respective
         fallbacks). These directories are created if necessary.

    In the cases of 1. and 3., the build script or executable is
    expected to be in the config dir.

    Additionally, the xmonad config binary and intermediate object files were
    moved to the cache directory (only relevant if using XDG or
    `XMONAD_CACHE_DIR`).

  * Added `Foldable`, `Functor`, and `Traversable` instances for `Stack`.

  * Fixed a bug when using multiple screens with different dimensions,
    causing some floating windows to be smaller/larger than the size they
    requested.

  * Added `Typeable layout` constraint to `LayoutClass`, making it possible to
    cast `Layout` back into a concrete type and extract current layout state
    from it.

  * Export constructor for `Choose` and `CLR` from `Module.Layout` to allow
    pattern-matching on the left and right sub-layouts of `Choose l r a`.

  * Compatibility with GHC 9.0

  * Document the help command in the help message.

  * Added `withUnfocused` function to `XMonad.Operations`, allowing for
    `X` operations to be applied to unfocused windows.

</details>

<details>
  <summary>New three-part structure</summary>

### Breaking Changes

  * Made `(<&&>)` and `(<||>)` non-strict in their right operand; i.e.,
    these operators now implement short-circuit evaluation so the right
    operand is evaluated only if the left operand does not suffice to
    determine the result.

  * Change `ScreenDetail` to a newtype and make `RationalRect` strict in
    its contents.

  * Added the `extensibleConf` field to `XConfig` which makes it easier for
    contrib modules to have composable configuration (custom hooks, …).

  * `util/GenerateManpage.hs` is no longer distributed in the tarball.
    Instead, the manpage source is regenerated and manpage rebuilt
    automatically in CI.

  * `DestroyWindowEvent` is now broadcasted to layouts to let them know
    window-specific resources can be discarded.

### Enhancements

  * Migrated `X.L.LayoutCombinators.(|||)` into `XMonad.Layout`,
    providing the ability to directly jump to a layout with the
    `JumpToLayout` message.

  * Recompilation now detects `stack.yaml` (can be a symlink) alongside
    `xmonad.hs` and switches to using `stack ghc`. We also updated INSTALL.md
    with instructions for cabal-install that lead to correct recompilation.

    Deprecation warnings during recompilation are no longer suppressed to make
    it easier for us to clean up the codebase. These can still be suppressed
    manually using an `OPTIONS_GHC` pragma with `-Wno-deprecations`.

  * Improve handling of XDG directories.

      1. If all three of xmonad's environment variables (`XMONAD_DATA_DIR,`
         `XMONAD_CONFIG_DIR`, and `XMONAD_CACHE_DIR`) are set, use them.
      2. If there is a build script called `build` (see
         [here](https://github.com/xmonad/xmonad-testing/tree/master/build-scripts)
         for usage examples) or configuration `xmonad.hs` in `~/.xmonad`, set
         all three directories to `~/.xmonad`.
      3. Otherwise, use the `xmonad` directory in `XDG_DATA_HOME`,
         `XDG_CONFIG_HOME`, and `XDG_CACHE_HOME` (or their respective
         fallbacks). These directories are created if necessary.

    In the cases of 1. and 3., the build script or executable is
    expected to be in the config dir.

    Additionally, the xmonad config binary and intermediate object files were
    moved to the cache directory (only relevant if using XDG or
    `XMONAD_CACHE_DIR`).

  * Added `Foldable`, `Functor`, and `Traversable` instances for `Stack`.

  * Added `Typeable layout` constraint to `LayoutClass`, making it possible to
    cast `Layout` back into a concrete type and extract current layout state
    from it.

  * Export constructor for `Choose` and `CLR` from `Module.Layout` to allow
    pattern-matching on the left and right sub-layouts of `Choose l r a`.

  * Added `withUnfocused` function to `XMonad.Operations`, allowing for
    `X` operations to be applied to unfocused windows.

### Bug Fixes

  * Fixed a bug when using multiple screens with different dimensions,
    causing some floating windows to be smaller/larger than the size they
    requested.

  * Compatibility with GHC 9.0

  * Document the help command in the help message.

</details>

liskin: > suppose what constitutes a breaking change and what's a bug fix can be bikeshedded for eternity

Indeed, so how about I write my concerns here and you then decide to consider them or not, and I will not interfere with that decision whatsoever, and then we consider it dealt with.

My primary concern is that most of the breaking changes aren't really breaking for the vast majority of configs, and them being mentioned prominently shadows stuff that's more important or more exciting. I'd almost argue that there are no real breaking changes at all: genmanpage is only breaking for distros, and the rest only breaks in special circumstances.

One way to deal with that concern is to give up and just reorder the changelog to have the important/exciting bits at the top without having any sections at all. Another that occurred to me just now and which may actually be good is to not give up, and since we're lucky that not a single breaking change needs to stay at the top, move the entire section to the end.

Oh and let's drop the `Document the help command in the help message` entry. We don't need to list every minor commit in there.

TheMC47:
> > ```
> > * [ ]  what about STYLE, CONFIG?
> > ```
> 
> 
> CONFIG is, imo, fully replaced by TUTORIAL.md.  STYLE still seems to be pretty accurate for the core at least, but should probably live inside CONTRIBUTING.md instead.  One thing I have a problem with is
> 
> 
> >

Regarding STYLE, we can move the relevant parts to `CONTRIBUTING.md` or `X.D.Developing`. Might be the right time to include a template for module headers

liskin:
> LGTM!

I'll merge this now and continue tracking the remaining TODOs in https://github.com/xmonad/xmonad/issues/341. Discussion can continue anywhere.

> I've been thinking a bit about whether that's a good thing or not.  Commit bits can be abused (as certain recent events have shown us), and with like 20 people having access to contrib already, are we just maximising our attack surface, waiting for something bad to happen?

I may be naive, but I wouldn't worry about that situation happening again: we do have a somewhat clearer governance structure now (still undocumented, but at least the overlap between org admins and people active in the project is better) and we've learned that being able to have honest real-time conversations is crucial—and that is exactly why I mentioned "communication with others" as one of the prerequisites.

Technically, though, you are indeed onto something. People with write/push access to xmonad-contrib can make releases (all the way to Hackage), and the more people have that access, the more it is likely one of them will have their credentials stolen and exploited by a third-party malicious actor. GitHub doesn't give us good tooling to protect from this kind of attack. We can move the release actions and hackage tokens to a separate repo that only core maintainers can trigger actions in, but this still leaves to option to create a tag/release on GitHub, and there doesn't seem to be any good way to protect that. I'm not too worried about us becoming a target of this kind of attack, though.

PR #345, opened 2021-10-29
ci: Improve release workflow

liskin:
#### [ci: Swap candidate/final release logic](../commit/cda4a52d3e28ebecdcc2bf13ace1d75348cf06c2)

During the release of xmonad 0.17.0, I realized that we need to be able to upload candidates before tagging the release on GitHub, because there might be issues with the tarball and Hackage may reject it. When that happened, I had to remove the release, delete the tag, upload the candidate manually to see what's wrong with it, try to fix it, upload it manually again, and so on.

This commit swaps the logic: when the workflow is invoked manually, it uploads the candidate. This can be done multiple times, and once everything is fine, the release can finally be tagged and it's released to Hackage proper. The only disadvantage is that we need to remember to try uploading the candidate. Not sure if there's a perfect solution…

#### [ci: Show error body from Hackage when it fails](../commit/58aae1875e7da3c3991e0daafdf8820821c387d5)

Prevents having to upload the candidate manually to see what's wrong.

### DIFF
--- a/.github/workflows/haskell-ci-hackage.patch
+++ b/.github/workflows/haskell-ci-hackage.patch
@@ -2,14 +2,16 @@ Piggy-back on the haskell-ci workflow for automatic releases to Hackage.
 
 This extends the workflow with two additional triggers:
 
- * When a release is created on GitHub, a candidate release is uploaded to
-   Hackage and docs are submitted for it as Hackage can't build them itself
-   (https://github.com/haskell/hackage-server/issues/925).
+ * When the Haskell-CI workflow is triggered manually with a non-empty version
+   input (matching the version in the cabal file), a candidate release is
+   uploaded to Hackage and docs are submitted for it as Hackage can't build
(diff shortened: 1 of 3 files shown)


liskin:
(Sorry for the crap title/description you got via e-mail, I wanted to see what `gh pr create --fill` does with multicommit PRs. Not surprisingly, it's rubbish.)

liskin:
Hm, and maybe this is all a waste of time as it seems Hackage is now able to build X11, xmonad and xmonad-contrib: https://hackage.haskell.org/package/xmonad-extras-0.17.0/reports/1
We never know when it breaks again, though. Makes me a bit sad :-/

slotThe:
This will work even now that Hackage is able to build X11, right?  In that case I'd still prefer this approach; having things not work when the release is already somewhat announced seems like something we don't need to go through a second time :)

liskin:
Yeah, I think I agree. I'm just disappointed about spending so much time on it.

PR #350, opened 2021-11-18
X.Operations: Make window borders opaque

Thiago4532:
### Description

The alpha wasn't set in the border, making the border color inconsistent when using transparent windows. This patch sets the border alpha to 0xff.

#### Before
image

#### After
image

### Checklist

  - [X] I've considered how to best test these changes (property, unit,
        manually, ...) and concluded: Open a transparent application and a non-transparent application (just like Firefox), now the border transparency is consistent between the two applications.

  - [N/A] I updated the `CHANGES.md` file

### DIFF
--- a/src/XMonad/Operations.hs
+++ b/src/XMonad/Operations.hs
@@ -250,7 +250,7 @@ setWindowBorderWithFallback :: Display -> Window -> String -> Pixel -> X ()
 setWindowBorderWithFallback dpy w color basic = io $
     C.handle fallback $ do
       wa <- getWindowAttributes dpy w
-      pixel <- color_pixel . fst <$> allocNamedColor dpy (wa_colormap wa) color
+      pixel <- setPixelSolid . color_pixel . fst <$> allocNamedColor dpy (wa_colormap wa) color
       setWindowBorder dpy w pixel
   where
     fallback :: C.SomeException -> IO ()
@@ -504,10 +504,14 @@ cleanMask km = do
     nlm <- gets numberlockMask
     return (complement (nlm .|. lockMask) .&. km)
 
+-- | Set the 'Pixel' alpha value to 255.
+setPixelSolid :: Pixel -> Pixel
+setPixelSolid p = (p .|. 0xff000000)
+
 -- | Get the 'Pixel' value for a named color.
 initColor :: Display -> String -> IO (Maybe Pixel)
 initColor dpy c = C.handle (\(C.SomeException _) -> return Nothing) $
-    (Just . color_pixel . fst) <$> allocNamedColor dpy colormap c
+    (Just . setPixelSolid . color_pixel . fst) <$> allocNamedColor dpy colormap c
     where colormap = defaultColormap dpy (defaultScreen dpy)
 
 ------------------------------------------------------------------------


slotThe (approved):
If others think this is a good change then let's merge it.  Just a small nit: `windows` in the commit message should probably be `window`

Thiago4532:
Is it possible to change the commit message? I've never done this before

geekosaur:
If you did this via command line, it's `git am -m "new commit message"`. If using Github's UI, you can click on the commit, then use the `Edit` button and change the commit message.

Thiago4532:
> If you did this via command line, it's `git am -m "new commit message"`. If using Github's UI, you can click on the commit, then use the `Edit` button and change the commit message.

Thanks

slotThe:
@Thiago4532 what do you use to set window transparency in your tests?  I just tried out this patch with `compton`s `inactive-opacity` option, but for me borders are still translucent

Thiago4532:
> @Thiago4532 what do you use to set window transparency in your tests? I just tried out this patch with `compton`s `inactive-opacity` option, but for me borders are still translucent

I've tested using alacritty and kitty terminal transparency using picom as the compositor. I'm not sure, but it looks like that compton/picom opacity rules does'nt respect the borders.

Thiago4532:
image
Alacritty and Kitty

slotThe:
Ah indeed, if I use alacritty's own transparency setting it seems to work (though it makes me question why alacritty has that option to begin with :))

Thiago4532:
> Ah indeed, if I use alacritty's own transparency setting it seems to work (though it makes me question why alacritty has that option to begin with :))

This option only makes the background transparent, rather than making everything transparent. You should use this option if you want the text to remain opaque.

slotThe:
Thanks!

liskin:
Hm, I wonder if we still need
https://github.com/xmonad/xmonad/blob/cde1a25bca9b7aeeb24af73588221f5f54ae770b/src/XMonad/Operations.hs#L267-L275
at all. It was introduced in https://github.com/xmonad/xmonad/pull/59 to fix https://code.google.com/archive/p/xmonad/issues/581 which seems exactly like the issue being fixed by this PR.

But! https://github.com/xmonad/xmonad/pull/59 only changed the code that updates the border between normal and focused, but forgot to update `setInitialProperties`:
https://github.com/xmonad/xmonad/blob/cde1a25bca9b7aeeb24af73588221f5f54ae770b/src/XMonad/Operations.hs#L299-L307
In most xmonad configs, new windows are focused and `setWindowBorderWithFallback` gets invoked immediately, so if `setWindowBorder` wasn't enough, it'd be somewhat difficult to notice. Still, if you change `normalBorderColor` and restart xmonad, it's just the `setWindowBorder` in `setInitialProperties` that handles changing the border color of all your inactive windows, not `setWindowBorderWithFallback`. Surely someone would notice if this set the wrong border color, right?

@geekosaur You understand this stuff more than me; is there a good reason to use the window's colormap instead of the default colormap? Does X11 start with a 24-bit colormap until a compositor is attached, perhaps? Or someone might use an indexed colormap or something?
Shall we use `setWindowBorderWithFallback` in `setInitialProperties` then? Or if this is truly unnecessary, drop `setWindowBorderWithFallback` as it's more expensive than `setWindowBorder`.

geekosaur:
There are still things that use 8 bit colormaps, notably older VNC-based protocols or low bandwidth mode, and using the root window 's colormap will do the wrong thing ("technicolor"). That's why the code was changed in the first place.

liskin:
> There are still things that use 8 bit colormaps, notably older VNC-based protocols or low bandwidth mode, and using the root window 's colormap will do the wrong thing ("technicolor"). That's why the code was changed in the first place.

Is there an easy way to test such an application? If my understanding is correct, simply restarting xmonad should still result in borders of all windows with non-TrueColor colormap having the wrong color. I'd love to reproduce that and possibly fix.

geekosaur:
The easiest way is to run an xmonad-based session in a VNC server in 8 bit mode.

PR #362, opened 2022-01-10
Fix nix flake compilation.

mikenrafter:
### Description

Was no longer compiling, attribute `gitIgnoreSource` didn't exist, it was moved to `lib.gitignoreSource`.

### Checklist

        manually, ...) and concluded: unit &/or manually.

Not sure that last checkbox is relevant here.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -10,7 +10,7 @@
       haskellPackages = prev.haskellPackages.override (old: {
         overrides = prev.lib.composeExtensions (old.overrides or (_: _: {}))
         (hself: hsuper: {
-          xmonad = hself.callCabal2nix "xmonad" (git-ignore-nix.gitIgnoreSource ./.) { };
+          xmonad = hself.callCabal2nix "xmonad" (git-ignore-nix.lib.gitignoreSource ./.) { };
         });
       });
     };


liskin:
Looks as if this broke because @IvanMalison synced [his fork](https://github.com/IvanMalison/gitignore.nix) of [gitignore.nix](https://github.com/hercules-ci/gitignore.nix) with upstream, and I'm only guessing as this seems the only meaningful explanation why it'd break, given last commit in both repos is a couple months ago but it broke recently (less than a wekk). Should we perhaps switch to the upstream directly so that problems like this are easier to diagnose and don't need the push history to explain?

TheMC47:
Thanks @mikenrafter for your contribution!

I know nothing about nix, but...

> Should we perhaps switch to the upstream directly so that problems like this are easier to diagnose and don't need the push history to explain?

Definitely, seems weird to rely on a fork that we don't control. 

@mikenrafter could you please take a look at the contributing guidelines, especially the one about [not merging the master branch into your branch](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md#rebasing-and-squashing-commits)? If this is confusing, just say so and we'll guide you through it

colonelpanic8:
> Looks as if this broke because @IvanMalison synced [his fork](https://github.com/IvanMalison/gitignore.nix) of [gitignore.nix](https://github.com/hercules-ci/gitignore.nix) with upstream, and I'm only guessing as this seems the only meaningful explanation why it'd break, given last commit in both repos is a couple months ago but it broke recently (less than a wekk). Should we perhaps switch to the upstream directly so that problems like this are easier to diagnose and don't need the push history to explain?

Yes we should move to using the hercules-ci flake. The only reason I pointed it at my fork at all was because hercules-ci did not have a flake at the time. Furthermore, if we add a flake.lock file we could avoid any issues with the upstream being changed in the future.

liskin:
> Furthermore, if we add a flake.lock file we could avoid any issues with the upstream being changed in the future.

Are there any downsides of this approach as well? In general we try to stay up to date with deps and the occasional CI failure and some amount of work to make everything compatible again is okay (and we do try to keep the amount of deps minimal to avoid having too much maintenance work). Is this thing likely to break again, and would freezing the version have no other negative consequences?

(Sorry for asking these stupid questions, I still haven't learned enough Nix to have answers myself.)

turboMaCk:
it's pretty much what you would expect from lock file. Anyway in this case the "dependecies" are gitignore integration for nix and flake utils. Don't think you risk loosing much in terms of updates if both already do all the things you need them to.

mikenrafter:
> not merging the master branch into your branch

Will do.

Issue #300, opened 2021-05-24
Add a willFloat function to detect whether a window will float in a ManageHook


andrea-berling:
### Problem Description

The problem is this: I would like for windows to grow in the opposite directions w.r.t. the default one (e.g. from left to right in Tabbed Simplest). To do this I used the `insertPosition` from xmonad-contrib. Now, I wanted to use it only for windows that do not float, as if I do it also for windows that do float what will happen is that an older window will stay on top of the newest, and thus some floating windows will remain hidden (it will look as if no new window was drawn at all). Thus I wanted to make an exception for floating windows in my ManageHook. Sadly `isFloat` from `XMonad.Hooks.RefocusLast` won't work, since when manageHook runs, the window isn't managed yet. Thus I wrote my own `willFloat` method based on the way XMonad decides if a window is floating or not. This does work in my manageHook and does achieve what I was aiming for. Maybe this will also be helpful to others, so I will leave the code for it below so that you can see if it may be a good method to be exported by xmonad core.

### Configuration File

```haskell
willFloat::Query Bool
willFloat = ask >>= \w -> liftX $ withDisplay $ \d -> do
  sh <- io $ getWMNormalHints d w
  let isFixedSize = isJust (sh_min_size sh) && sh_min_size sh == sh_max_size sh
  isTransient <- isJust <$> io (getTransientForHint d w)
  return (isFixedSize || isTransient)
```
Example configuration:
```haskell
module Main (main) where

import XMonad
import Data.Maybe(isJust)
import XMonad.Hooks.InsertPosition(insertPosition,Focus(..),Position(..))

willFloat::Query Bool
willFloat = ask >>= \w -> liftX $ withDisplay $ \d -> do
  sh <- io $ getWMNormalHints d w
  let isFixedSize = isJust (sh_min_size sh) && sh_min_size sh == sh_max_size sh
  isTransient <- isJust <$> io (getTransientForHint d w)
  return (isFixedSize || isTransient)

main :: IO ()
main = xmonad def
        { modMask = mod4Mask -- Use Super instead of Alt
        , terminal = "urxvt"
        , manageHook = fmap not willFloat --> insertPosition Below Newer
        }
```

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - [x] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

kopr12:
[...]

PR #371, opened 2022-01-22
Add function to detect floating windows in ManageHook

andrea-berling:
This PR adds a function to `ManageHooks` to check whether the window to be managed is going to be a floating window or not. This feature and the problem it addresses are described more in detail in issue #300. An example configuration and an example use case are described in the issue. Under suggestion of @slotThe I decided to turn that issue into an actual PR for xmonad core. I've run the added code against the unit tests and nothing broke. I also made sure to test it manually using Xephyr to confirm that it was in fact doing what it tries to achieve.

### Checklist

        manually, ...) and concluded: unit testing + manual testing

### DIFF
--- a/src/XMonad/ManageHook.hs
+++ b/src/XMonad/ManageHook.hs
@@ -106,6 +106,14 @@ getStringProperty d w p = do
   md <- io $ getWindowProperty8 d a w
   return $ fmap (map (toEnum . fromIntegral)) md
 
+-- | Return whether the window will be a floating window or not
+willFloat :: Query Bool
+willFloat = ask >>= \w -> liftX $ withDisplay $ \d -> do
+  sh <- io $ getWMNormalHints d w
+  let isFixedSize = isJust (sh_min_size sh) && sh_min_size sh == sh_max_size sh
+  isTransient <- isJust <$> io (getTransientForHint d w)
+  return (isFixedSize || isTransient)
+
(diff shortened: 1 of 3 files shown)


src/XMonad/ManageHook.hs
```diff
@@ -106,6 +106,14 @@ getStringProperty d w p = do
   md <- io $ getWindowProperty8 d a w
   return $ fmap (map (toEnum . fromIntegral)) md
 
+-- | Return whether the window will be a floating window or not
+willFloat :: Query Bool
+willFloat = ask >>= \w -> liftX $ withDisplay $ \d -> do
```
geekosaur: Would it make more sense to refactor the code in XMonad.Operations to separate out its check, and then use that here?

geekosaur (approved):
Aside from my question above (which is not critical), this looks okay to me.

src/XMonad/ManageHook.hs
andrea-berling: It totally does, I did just that in my last commit. Thanks for pointing it out, I completely missed it the first time around :slightly_smiling_face:

geekosaur (approved):
LGTM!

slotThe (approved):
LGTM as well!  I will merge in a few if no one else objects

slotThe:
Thank you!

PR #374, opened 2022-02-10
Add stripModMask to customize cleanMask/extraModifiers

### LINKED ISSUE #172: Key bindings ignore numlock state
### Problem Description

I've got a keyboard which uses some keys on the numpad as volume up / volume down / mute keys if numlock is off, and normal numbers and operators if numlock is on. I'm trying to bind them so my volume up / down / mute keys actually work, but xmonad ignores numlock state, so my key bindings always trigger, even if numlock is on, which means I can't use my numpad.

I know xmonad 0.9 used to have a `numlockMask` config option which could disable this behavior, but it got deprecated and I guess removed since it doesn't work anymore: https://wiki.haskell.org/Xmonad/Config_archive/Template_xmonad.hs_(0.9). I see the current code refers to `numberlockMask` in some places and numlock is ignored because of `cleanMask` in this place: https://github.com/xmonad/xmonad/blob/master/src/XMonad/Main.hs#L295, is it even possible to use them to fix this problem? I can't find anyone using those in their config file and I'm too new to Haskell to know if I can somehow override them or write my own function to handle key bindings bypassing XMonad's one. I've been reading the docs on hackage for days, but can't find anything helpful.

Ideally I'd like to disable ignoring numlock key only for those key bindings which happen to use the numpad on my keyboard, but disabling it for all key bindings is a viable solution too since I would still be able to make both work.

I'm using xmonad 0.13 from Arch' repo, I tried to fiddle with the code from `master` and rebuild it from source (which is still an ugly solution, but better than nothing) but cabal fails to configure it due to "Encountered missing dependencies: semigroups -any" and I don't know how to solve it either, installing `semigroups`, whatever that is, manually doesn't work, so nope, I didn't check it on code from `master` nor `xmonad-testing`.

### Configuration File

```haskell
import XMonad
import System.Exit
import Graphics.X11.ExtraTypes.XF86

import qualified Data.Map as M

main = xmonad def
	{ modMask = mod4Mask
	, keys    = cKeyBindings }

cKeyBindings conf@(XConfig {XMonad.modMask = modm}) = M.fromList $
	[ ((modm .|. shiftMask, xK_Return), spawn $ XMonad.terminal conf)
	, ((modm,               xK_p     ), spawn "dmenu_run")
	, ((modm .|. shiftMask, xK_Escape), io (exitWith ExitSuccess))
	, ((modm              , xK_Escape), spawn "xmonad --recompile; xmonad --restart")
	, ((0                 , xF86XK_AudioMute), spawn "xmessage works") ]
```

### Checklist

  - [ ] I tested my configuration with [xmonad-testing](https://github.com/xmonad/xmonad-testing)

liskin:
### Description

Added `stripModMask` to allow customizing which modifiers are irrelevant for key bindings. Useful for binding numpad keys only when Num Lock is off, or to make Mod5 irrelevant in addition to the default Num/Caps Lock.

Fixes: https://github.com/xmonad/xmonad/issues/172
Relates: https://github.com/xmonad/xmonad-contrib/issues/290

### Checklist

  - [X] I've considered how to best test these changes (property, unit, manually, ...) and concluded: I'm hoping that someone who requested this in https://github.com/xmonad/xmonad/issues/172 will test it manually :-)

### DIFF
--- a/src/XMonad/Config.hs
+++ b/src/XMonad/Config.hs
@@ -41,6 +41,7 @@ import qualified XMonad.StackSet as W
 import Control.Monad.State (gets)
 import Data.Bits ((.|.))
 import Data.Default.Class
+import Data.Functor ((<&>))
 import Data.Monoid
 import qualified Data.Map as M
 import System.Exit
@@ -72,8 +73,8 @@ defaultModMask = mod1Mask
 -- this if you wish to only strip Caps Lock; e.g., if you need to bind numpad keys
 -- but only when Num Lock is off (or on). Another use case is adding
 -- 'mod5Mask' to the list of stripped/irrelevant modifiers.
-defaultStripModMask :: X KeyMask
-defaultStripModMask = gets ((lockMask .|.) . numberlockMask)
+defaultStripModMask :: X [KeyMask]
+defaultStripModMask = gets numberlockMask <&> \numLockMask -> [lockMask, numLockMask]
 
 -- | Width of the window border in pixels.
 --
--- a/src/XMonad/Config.hs
+++ b/src/XMonad/Config.hs
@@ -38,6 +38,7 @@ import XMonad.Layout
 import XMonad.Operations
 import XMonad.ManageHook
 import qualified XMonad.StackSet as W
+import Control.Monad.State (gets)
 import Data.Bits ((.|.))
 import Data.Default.Class
 import Data.Monoid
@@ -58,13 +59,21 @@ import Graphics.X11.Xlib.Extras
 workspaces :: [WorkspaceId]
 workspaces = map show [1 .. 9 :: Int]
 
--- | modMask lets you specify which modkey you want to use. The default
+-- | 'modMask' lets you specify which modkey you want to use. The default
 -- is mod1Mask ("left alt").  You may also consider using mod3Mask
 -- ("right alt"), which does not conflict with emacs keybindings. The
 -- "windows key" is usually mod4Mask.
 --
 defaultModMask :: KeyMask
 defaultModMask = mod1Mask
+
+-- | 'stripModMask' lets you specify which modifiers are irrelevant for key
+-- bindings. The default is Num Lock and Caps Lock. You will need to override
+-- this if you wish to only strip Caps Lock if you need to bind numpad keys
+-- but only when Num Lock is off (or on). Another use case is adding
+-- 'mod5Mask' to the list of stripped/irrelevant modifiers.
+defaultStripModMask :: X KeyMask
+defaultStripModMask = gets ((lockMask .|.) . numberlockMask)
 
 -- | Width of the window border in pixels.
 --
@@ -264,6 +273,7 @@ instance (a ~ Choose Tall (Choose (Mirror Tall) Full)) => Default (XConfig a) wh
     , XMonad.normalBorderColor  = normalBorderColor
     , XMonad.focusedBorderColor = focusedBorderColor
     , XMonad.modMask            = defaultModMask
+    , XMonad.stripModMask       = defaultStripModMask
     , XMonad.keys               = keys
     , XMonad.logHook            = logHook
     , XMonad.startupHook        = startupHook
--- a/src/XMonad/Config.hs
+++ b/src/XMonad/Config.hs
@@ -69,7 +69,7 @@ defaultModMask = mod1Mask
 
 -- | 'stripModMask' lets you specify which modifiers are irrelevant for key
 -- bindings. The default is Num Lock and Caps Lock. You will need to override
--- this if you wish to only strip Caps Lock if you need to bind numpad keys
+-- this if you wish to only strip Caps Lock; e.g., if you need to bind numpad keys
 -- but only when Num Lock is off (or on). Another use case is adding
 -- 'mod5Mask' to the list of stripped/irrelevant modifiers.
 defaultStripModMask :: X KeyMask
--- a/src/XMonad/Operations.hs
+++ b/src/XMonad/Operations.hs
@@ -56,8 +56,8 @@ import qualified XMonad.StackSet as W
 
 import Data.Maybe
 import Data.Monoid          (Endo(..),Any(..))
-import Data.List            (nub, (\\), find)
-import Data.Bits            ((.|.), (.&.), complement, testBit)
+import Data.List            (nub, (\\), find, foldl', subsequences)
+import Data.Bits            ((.|.), (.&.), complement, bit, testBit, clearBit, countTrailingZeros)
 import Data.Function        (on)
 import Data.Ratio
 import qualified Data.Map as M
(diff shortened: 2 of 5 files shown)


src/XMonad/Config.hs
```diff
@@ -58,14 +59,22 @@ import Graphics.X11.Xlib.Extras
...
 -- ("right alt"), which does not conflict with emacs keybindings. The
 -- "windows key" is usually mod4Mask.
 --
 defaultModMask :: KeyMask
 defaultModMask = mod1Mask
 
+-- | 'stripModMask' lets you specify which modifiers are irrelevant for key
+-- bindings. The default is Num Lock and Caps Lock. You will need to override
+-- this if you wish to only strip Caps Lock if you need to bind numpad keys
+-- but only when Num Lock is off (or on). Another use case is adding
```
slotThe: This is worded a bit weirdly (to my non-native ears :), I would suggest adding something to combine these dependent clauses like

```suggestion
-- bindings. The default is Num Lock and Caps Lock. You will need to override
-- this if you wish to only strip Caps Lock; e.g., if you need to bind numpad keys
-- but only when Num Lock is off (or on). Another use case is adding
```

YMMV on the punctuation (I think this is the correct way to do it, but it looks horrible so ¯\\\_(ツ)\_/¯)

src/XMonad/Operations.hs
```diff
@@ -500,17 +500,20 @@ isClient :: Window -> X Bool
...
+-- (by default numlock and capslock, can be overridden in 'stripModMask')
 extraModifiers :: X [KeyMask]
 extraModifiers = do
-    nlm <- gets numberlockMask
-    return [0, nlm, lockMask, nlm .|. lockMask ]
+    smm <- join $ asks $ stripModMask . config
+    return $ map (foldl' (.|.) 0) (subsequences (bits smm))
+  where
+    bits 0 = []
+    bits n = let b = countTrailingZeros n in bit b : bits (n `clearBit` b)
```
slotThe: This would be much more understandable if we just specified a list in `stripModMask`, right?[^1]  Is there any reason why we don't do that and instead or everything together there already?

[^1]: AFAICT we could then just use `subsequences` alone.

slotThe (commented):
Pinging @kurogetsusai and @ibbem for testing

src/XMonad/Operations.hs
liskin: "Understandable" refers to the implementation or the user interface? I wanted to abstract away the implementation detail that we need to grab all bit combinations and just expose a _mask_ that's directly used in `cleanMask` and processed into combinations in here.

You could argue, however, that abstracting the implementation detail is harmful in here, because people who want to not strip numlock will need to manually grab some keys with and without numlock, so abstracting it away will make it harder for them to get the right mental models. Hm, not sure what to do. :-/

slotThe: I would argue that the fact that key masks are or'd together is an implementation detail in itself, which is not necessarily known to the user (many people don't know what the funnly looking `.|.` is doing in their configs :).  Intuitively, I would expect an option along the lines of "filter out these modifiers" to take a list of key masks instead of a single one

liskin: Okay, that makes sense: 67304e9

src/XMonad/Config.hs
```diff
@@ -58,14 +60,22 @@ import Graphics.X11.Xlib.Extras
...
 defaultModMask :: KeyMask
 defaultModMask = mod1Mask
 
+-- | 'stripModMask' lets you specify which modifiers are irrelevant for key
+-- bindings. The default is Num Lock and Caps Lock. You will need to override
+-- this if you wish to only strip Caps Lock; e.g., if you need to bind numpad keys
+-- but only when Num Lock is off (or on). Another use case is adding
+-- 'mod5Mask' to the list of stripped/irrelevant modifiers.
+defaultStripModMask :: X [KeyMask]
+defaultStripModMask = gets numberlockMask <&> \numLockMask -> [lockMask, numLockMask]
```
liskin, line 77: Is `gets ((:[lockMask]) . numberlockMask)` better? I don't know. :-)

slotThe: I don't think one is more obscure than the other, so feel free to pick the one you like best :)

slotThe (approved):

ibbem:
I currently try to test this change, but while updating my configuration I'm stuck with trying to figure out how to bind a key while numlock is active. The problem is, that the numlock mask is determined during the startup of xmonad, so it is not available at configuration time. (I also tried to run the numlock extraction code before the configuration code, but for that I would have to open a X connection.)
This is how my minimal attempt looks like:
```haskell
import XMonad
import XMonad.Util.CustomKeys

main :: IO ()
main =
  xmonad
    def
      { stripModMask = lockMask,
        keys =
          customKeys (const []) $
            \XConfig {modMask = modMask} ->
              [ ((modMask, xK_y), xmessage "without numlock"),
                ((modMask .|. undefined {- how do I obtain the numlock mask? -}, xK_y), xmessage "with numlock")
              ]
      }
```
Is it possible to use a static numlock mask? When does the numlock mask actually change (I saw that it is updated when a `MappingNotifyEvent` is handled)?

When I use different mappings for the caps lock state it works as expected with this configuration:
```haskell
import XMonad
import XMonad.Util.CustomKeys

main :: IO ()
main =
  xmonad
    def
      { stripModMask = gets numberlockMask,
        keys =
          customKeys (const []) $
            \XConfig {modMask = modMask} ->
              [ ((modMask, xK_y), xmessage "without caps lock"),
                ((modMask .|. lockMask, xK_y), xmessage "with caps lock")
              ]
      }
```
So the actual logic of this change seems to work.

---

For the interface it would probably be a good idea to offer a function (maybe in xmonad-contrib) to ignore a modifier for all other mappings. I'm not exactly sure about the interface for such a function, some quick ideas include:

```haskell
keysWithoutIgnoringModifiers1 :: KeyMask -> [((KeyMask, KeySym), X ())] -> XConfig l -> XConfig l
```
This would automatically remove the modifiers from the stripModMask, duplicate the key mappings from the given config, with and without the modifier, and add the given key mapping unmodified.

```haskell
keysWithoutIgnoringModifiers2 :: KeyMask -> (XConfig Layout -> [((KeyMask, KeySym), X ())]) -> XConfig l -> XConfig l
```
Variation on the above that use a function like the one used for `keys` or `customKeys`.

```haskell
ignoreModifiers :: KeyMask -> [((KeyMask, KeySym), X ())] -> [((KeyMask, KeySym), X ())]
```
This would only duplicate the key mappings with and without the modifier. Note that it is unnecessary to adapt this one to use a function like `keys` or `customKeys` do, as its intended use case already allows such a function to be used (see example below).

Usage of these would look like:
```haskell
import XMonad
import XMonad.Util.CustomKeys

main :: IO ()
main =
  xmonad
    $ keysWithoutIgnoringModifiers2 lockMask (\XConfig {modMask = modMask} ->
       [ ((modMask, xK_y), xmessage "without caps lock"),
         ((modMask .|. lockMask, xK_y), xmessage "with caps lock")
       ])
    $ def
      { keys =
          customKeys (const []) $
            \XConfig {modMask = modMask} ->
              [ {- some custom keys -} ]
      }
```
or
```haskell
import XMonad
import XMonad.Util.CustomKeys

main :: IO ()
main =
  xmonad
    def
      { stripModMask = gets numberlockMask,
        keys =
          customKeys (const []) $
            \XConfig {modMask = modMask, workspaces = workspaces} ->
              ignoreModifiers lockMask
                [ {- some other key bindings -} ]
              ++
                [ ((modMask, xK_y), xmessage "without caps lock"),
                  ((modMask .|. lockMask, xK_y), xmessage "with caps lock")
                ]
      }
```

I think I prefer `keysWithoutIgnoringModifiers2`, although there is probably a better name for it.

PR #375, opened 2022-02-16
Cabal build support

geekosaur:
### Description

Add support for cabal builds based on presence of a `cabal.project` file, analogous to our support for stack builds based on `stack.yaml`.

### Checklist

        manually, ...) and concluded: needs manual testing

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,9 @@
 
 ### Enhancements
 
+  * Added support for automated builds using cabal, if a `cabal.project`
+    file is present. This mirrors the existing stack support.
+
   * Added custom cursor shapes for resizing and moving windows.
 
 ### Bug Fixes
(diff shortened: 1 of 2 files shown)


geekosaur:
I need to do a bit more testing, since this overrides the `INSTALL.md` workflow while adding a new one.

geekosaur:
I talked about this a bit on IRC, but I'm expanding on it and recording it here.

I'm going to abandon the `cabal.project`-driven implementation, since the existing `cabal install --lib` stuff already uses it (and must) in the case where `xmonad` and `xmonad-contrib` are to be installed from git. Instead, I want to use an `xmonad.cabal` file, which enables the use of additional dependencies. It does mean the file must be more complete than a `stack.yaml`, but that is unavoidable: `stack` and `cabal` have different divisions of labor between their config files vs. their cabal files. But it solves the problem that it is impossible to have additional dependencies (`dbus`, `utf8-string`, etc.) without vendoring them in `cabal.project` with the original implementation.

One open question is whether to stick to a specific cabal file as described above, or do what `cabal` does and search for any `*.cabal` file in the build directory. My original proposal goes for the simple solution, but @liskin said he found this odd.

sol:
https://github.com/xmonad/xmonad/pull/523 does not mandate any specific name for the `.cabal` file + `cabal.project` is transparently supported.

slotThe:
I guess we can close this by way of #523

PR #390, opened 2022-04-18
Document sample build scripts for stack and cabal

geekosaur:
This commit depends on https://github.com/xmonad/xmonad-contrib/pull/710.

### DIFF
--- a/src/XMonad/StackSet.hs
+++ b/src/XMonad/StackSet.hs
@@ -94,21 +94,20 @@ import qualified Data.Map  as M (Map,insert,delete,empty)
 --    resulting data structure will share as much of its components with
 --    the old structure as possible.
 --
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -356,6 +356,10 @@ exec stack ghc --  \
 
 Don't forget to mark the file as `+x`: `chmod +x build`!
 
+Some example build scripts for `stack` and `cabal` are provided in the
+`scripts/build` directory of the `xmonad-contrib` distribution. You can
+see those online [here][].
```
liskin: "here" is not a very descriptive link name, so it's better to just include the URI inline.

geekosaur: Yeh, I'm also thinking I should have included some reasons one might want to use them.

geekosaur:
I think I just messed this up

PR #401, opened 2022-06-17
Lint some pieces of code

unsafe-andrew-old:
Perform some lints, improve readability

### DIFF
--- a/src/XMonad/Main.hs
+++ b/src/XMonad/Main.hs
@@ -87,14 +87,12 @@ usage :: IO ()
 usage = do
     self <- getProgName
     putStr . unlines $
-        concat ["Usage: ", self, " [OPTION]"] :
-        "Options:" :
-        "  --help                       Print this message" :
-        "  --version                    Print the version number" :
-        "  --recompile                  Recompile your xmonad.hs" :
-        "  --replace                    Replace the running window manager with xmonad" :
-        "  --restart                    Request a running xmonad process to restart" :
-        []
+         [concat ["Usage: ", self, " [OPTION]"], "Options:",
+         "  --help                       Print this message",
+         "  --version                    Print the version number",
+         "  --recompile                  Recompile your xmonad.hs",
+         "  --replace                    Replace the running window manager with xmonad",
+         "  --restart                    Request a running xmonad process to restart"]
 
 -- | Build the xmonad configuration file with ghc, then execute it.
 -- If there are no errors, this function does not return.  An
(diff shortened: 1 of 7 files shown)


src/XMonad/Main.hs
```diff
@@ -87,14 +87,12 @@ usage :: IO ()
...
     putStr . unlines $
-        concat ["Usage: ", self, " [OPTION]"] :
-        "Options:" :
-        "  --help                       Print this message" :
-        "  --version                    Print the version number" :
-        "  --recompile                  Recompile your xmonad.hs" :
-        "  --replace                    Replace the running window manager with xmonad" :
-        "  --restart                    Request a running xmonad process to restart" :
-        []
+         [concat ["Usage: ", self, " [OPTION]"], "Options:",
```
liskin, line 90: This would probably be nicer this way:

```haskell
[ "Usage: " <> self <> " [OPTION]"
, "Options:"
, "  --help …"
, …
]
```
```

**Review by liskin — COMMENTED** [review]
Looks all right. Does this address all hlint warnings or just a few select ones?

Would you perhaps want to help us enable hlint in the CI like we do in xmonad-contrib? (We'd be happy to do that ourselves sooner or later if not.)

**Comment by TheMC47** [comment]
Thank you for the PR. I'm a bit on the fence about this, I'd want to setup an hlint action first and discuss the issues before we merge this.

File: src/XMonad/ManageHook.hs [hunk 2]
```diff
@@ -65,7 +65,7 @@ infixr 3 <&&>, <||>
 
 -- | '||' lifted to a 'Monad'.
 (<||>) :: Monad m => m Bool -> m Bool -> m Bool
-(<||>) x y = ifM x (pure True) y
+(<||>) x = ifM x (pure True)
```
TheMC47, line 68: This feels a bit weird

slotThe: should probably transform this to `x <||> y = ifM x (pure True) y` instead

unsafe-andrew-old:
> Thank you for the PR. I'm a bit on the fence about this, I'd want to setup an hlint action first and discuss the issues before we merge this.

Okay, should I create another PR to fix those two pcs of code?

unsafe-andrew-old:
> Looks all right. Does this address all hlint warnings or just a few select ones?

Only except for unused language pragmas

liskin:
> Okay, should I create another PR to fix those two pcs of code?

You can just force-push into this one.

liskin:
I amended your commit with https://github.com/xmonad/xmonad/pull/401#discussion_r900369394, enabled hlint in CI and fixed all the remaining hints and merged. Thanks!

liskin:
We did break the Nix build, however, so I started a conversation about this here: https://github.com/NixOS/nixpkgs/commit/36d5761b3e5aca9742fec85107e3d308a9af872c#commitcomment-76979989

PR #402, opened 2022-06-23
Add an hlint action

unsafe-andrew-old:


slotThe:
Related: https://github.com/xmonad/xmonad-contrib/issues/669

liskin:
Yeah, I've been looking at this since yesterday: https://github.com/liskin/xmonad/actions?query=branch%3Ahlint

liskin:
@andrei-lus The GitHub Action you suggested here seemed unmaintained, so I just did it the same way it's done in xmonad-contrib for consistency. As @slotThe mentioned, a separate action has been considered and it's something we'd possibly like to explore, but this wasn't the right one.

Issue #389, opened 2022-04-15
internal error: evacuate: strange closure type 4721 (Xmonad with 9.2.2 on ubuntu)


Abastro:
### Problem Description

When I set up a simple test xmonad instance on GHC 9.2.2, it crashes often (usually when I press a key).

I got the following in the error log (took time to catch this one).

```
xmonad-x86_64-linux: internal error: evacuate: strange closure type 4721
    (GHC version 9.2.2 for x86_64_unknown_linux)
    Please report this as a GHC bug:  https://www.haskell.org/ghc/reportabug
```

(Xmonad is quite a complex project to report a bug on GHC though)

### Steps to Reproduce

Quite unpredictable, but usually when:
1. Run xmonad with a desktop file in /usr/share/xsessions on ubuntu
2. Press Super+D or Super+Q to have xmonad do some "action"
3. Crashes occasionally

### Configuration File

```haskell
module Main where

import XMonad
import XMonad.Config.Desktop
import XMonad.Config.Gnome
import XMonad.Hooks.EwmhDesktops
import XMonad.Hooks.SetWMName
import XMonad.Util.EZConfig
import XMonad.Util.Run

main :: IO ()
main = do
  --dirs <- getDirectories
  --let xmDir = cfgDir dirs
  spawn "feh --bg-scale $HOME/Pictures/Universe/Andromeda_Galaxy.jpg"

  xmonad . ewmhFullscreen $
    desktopConfig
      { modMask = mod4Mask,
        startupHook = startupHook desktopConfig <> do
          setWMName "LG3D"
          gnomeRegister, -- Registers xmonad with gnome
        terminal = "gnome-terminal"
        --manageHook =  isDialog --> doFloat
      }
      `additionalKeysP` [("M-d", safeSpawnProg "nautilus")]

```

### Checklist

  - [x] I've read [CONTRIBUTING.md](https://github.com/xmonad/xmonad/blob/master/CONTRIBUTING.md)

  - I tested my configuration
    - [x] With `xmonad` version 0.17.0
    - [x] With `xmonad-contrib` version 0.17.0

geekosaur:
For the record, we're trying it with a minimal config (`main = xmonad def`) to see if it still happens. I can't say I expect much to change, though.

Also for the record, I ran most of a day with [my config](https://github.com/geekosaur/xmonad.hs/blob/skkukuk/xmonad.hs) under 9.2.2 and couldn't reproduce it.

[...]

PR #404, opened 2022-07-06
attempt to work around the join point bug (#389)

geekosaur:
### Description

Try rephrasing dons's happy little HOF as a mere tail recursion to avoid ghc #21708.

### Checklist

        manually, ...) and concluded: manually

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -13,6 +13,13 @@
   * Fixed border color of windows with alpha channel. Now all windows have the
     same opaque border color.
 
+  * Change the main loop to try to avoid [GHC bug 21708] on systems
+    running GHC 9.2 up to version 9.2.3. The issue has been fixed in
+    [GHC 9.2.4] and all later releases.
+
+[GHC bug 21708]: https://gitlab.haskell.org/ghc/ghc/-/issues/21708
+[GHC 9.2.4]: https://discourse.haskell.org/t/ghc-9-2-4-released/4851
+
 ## 0.17.0 (October 27, 2021)
 
 ### Enhancements
--- a/src/XMonad/Main.hs
+++ b/src/XMonad/Main.hs
@@ -267,10 +267,11 @@ launch initxmc drs = do
             userCode $ startupHook initxmc
 
             rrData <- io $ xrrQueryExtension dpy
-            let rrUpdate = when (isJust rrData) . void . xrrUpdateConfiguration
 
             -- main loop, for all you HOF/recursion fans out there.
-            forever $ prehandle =<< io (nextEvent dpy e >> rrUpdate e >> getEvent e)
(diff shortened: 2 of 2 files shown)


geekosaur:
not so easy, sadly. and to answer a question in IRC from @liskin, anything that does the same thing as `forever` will also hit the Cmm optimizer bug and end up trashing the heap.

geekosaur:
And once I got enough pieces moved for it all to compile, it works (on 8.10.7; I was never able to reproduce the 9.x bug for some reason). We need someone who did reproduce it to try. @Abastro? (This is a workaround; the fix is a release blocker for ghc 9.4, since it's a code generator bug.)

geekosaur:
Just adding a changelog entry, which I should have done to begin with.

CHANGES.md
```diff
@@ -11,6 +11,12 @@
   * Fixed border color of windows with alpha channel. Now all windows have the
     same opaque border color.
 
+  * Change the main loop to try to avoid [GHC bug 21708] on systems running
+    GHC 9.0 or later. A fixed GHC is forthcoming, as the bug is a blocker for
+    GHC 9.4.1.
+
+[GHC bug 21708]: https://gitlab.haskell.org/ghc/ghc/-/issues/21708
```
slotThe: Afaik the bug is specifically with 9.2.x (I can't find anyone being able to reproduce it with 9.0.x).  At this point, we might also mention that it's fixed with [9.2.4](https://discourse.haskell.org/t/ghc-9-2-4-released/4851)

src/XMonad/Main.hs
```diff
@@ -266,10 +266,11 @@ launch initxmc drs = do
             userCode $ startupHook initxmc
 
             rrData <- io $ xrrQueryExtension dpy
-            let rrUpdate = when (isJust rrData) . void . xrrUpdateConfiguration
 
             -- main loop, for all you HOF/recursion fans out there.
-            forever $ prehandle =<< io (nextEvent dpy e >> rrUpdate e >> getEvent e)
+            -- forever $ prehandle =<< io (nextEvent dpy e >> rrUpdate e >> getEvent e)
+            -- sadly, 9.x join points mishandle the above and trash the heap (see #389)
```
slotThe:
```suggestion
            -- sadly, 9.2.x join points mishandle the above and trash the heap (see #389)
```

slotThe (approved):
While not having heard back from @Abastro, in https://github.com/xmonad/xmonad/issues/389 we had @tbidne and @saep confirm that this fix worked for them.  I think it's pretty safe to merge this at this point.

tbidne:
@slotThe Indeed, I have not had any crashes since switching to the `forever-away` branch.

PR #405, opened 2022-07-08
X.Operations: Export setNumlockMask, grabKeys

slotThe:
### Description

As discussed in xmonad/xmonad-contrib/#703, certain functions that
X.U.Grab has vendored should really just be exported from core.

Related: https://github.com/xmonad/xmonad-contrib/pull/703

---

I'm not sure about a contrib PR right now, since we'd then again have to depend on xmonad 0.17.1 (let's say) for contrib 0.17.1; might not be a good look for a minor version bump?  Probably best to wait for 0.18.0 or something (opinions welcome, though)

---

### Checklist

        manually, ...) and concluded: tested manually (and am running this right now); seems fine.

### DIFF
--- a/src/XMonad/Operations.hs
+++ b/src/XMonad/Operations.hs
@@ -33,7 +33,7 @@ module XMonad.Operations (
     -- * Keyboard and Mouse
     cleanMask, extraModifiers,
     mouseDrag, mouseMoveWindow, mouseResizeWindow,
-    setButtonGrab, setFocusX,
+    setButtonGrab, setFocusX, cacheNumlockMask, mkGrabs,
 
     -- * Messages
     sendMessage, broadcastMessage, sendMessageWithNoRefresh,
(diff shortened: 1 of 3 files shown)


src/XMonad/Operations.hs
```diff
@@ -438,6 +438,40 @@ setFocusX w = withWindowSet $ \ws -> do
                   currentTime
               timedEvents = [ keyPress, keyRelease, buttonPress, buttonRelease, enterNotify, leaveNotify, selectionRequest ]
 
+setNumlockMask :: X ()
+setNumlockMask = do
```
liskin: Now that we're exporting this, maybe we could come up with a better name. Seems to me this is more of a "updateNumlockMask" or "refreshNumlockMask" or "cacheNumlockMask" than a "set" :-)

(And I wish we had a way to do handleEventHooks which ran after the core has handled the event, because then we wouln't need to export this and [customRegrabEvHook](https://github.com/TheMC47/xmonad-contrib/blob/33d1f2eccf5e606d977e5960a7b5d35e5609b1bf/XMonad/Util/Grab.hs#L87) didn't have to duplicate everything that the core handler does.)

liskin (approved):
Don't have an opinion on the version bounds yet.

src/XMonad/Operations.hs
slotThe: Good point!  I like `cacheNumlockMask` the best, I think.

> (And I wish we had a way to do handleEventHooks which ran after the core has handled the event, because then we wouln't need to export this and customRegrabEvHook didn't have to duplicate everything that the core handler does.)

Indeed :(

liskin: There isn't anything really stopping us from adding more hooks to core… apart from this feeling that if nobody added them by now surely they can't really be that useful. 😆

PR #406, opened 2022-07-09
man/Makefile: Add a clean target

felixonmars:
### Description

Add a clean target to ease cleanup of generated files.

### Checklist

### DIFF
--- a/man/Makefile
+++ b/man/Makefile
@@ -9,3 +9,6 @@ xmonad.1: xmonad.1.markdown
 
 xmonad.1.html: xmonad.1.markdown
 	pandoc --from=markdown --to=html --standalone --table-of-contents --output=$@ $<
+
+clean:
+	rm xmonad.1.markdown xmonad.1 xmonad.1.html


liskin:
The Makefile isn't intended to be used by end-users or even distributors (but it should say that!, and I'll update this later), it's our internal tooling to *update* the generated files, which are committed to git and shipped in the tarball.

So I wonder, what's the use case for deleting these files?

felixonmars:
> what's the use case for deleting these files?

As a distribution packager, we usually prefer to avoid using any pre-generated files as possible, because the ability to re-construct artifacts from the real source format is important to us.

It's totally fine if this Makefile is not meant to be stable and subjected to change at any time, we can adapt to it or probably just remove the generated files manually like what is done now.

liskin:
I see. What about `make --always-make` then? That also regenerates all the files and doesn't require a `clean` target listing all the generated files; and it's (almost) exactly what we run in our CI.

felixonmars:
That works too! Thanks :)

PR #407, opened 2022-07-12
TUTORIAL: "Fix" links to xmobar

liskin:
### Description

xmobar moved to codeberg :-/
And yeah, linking to `#headline-number` will be fun to maintain…

### Checklist

  - [X] I've considered how to best test these changes (property, unit, manually, ...) and concluded: checked manually and it'll be linkchecked once xmonad-web pulls it in its nightly CI

  - N/A I updated the `CHANGES.md` file

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -592,14 +592,14 @@ Config { overrideRedirect = False
 ```
 
 First, we set the font to use for the bar, as well as the colors.  The
-position options are documented well on the [xmobar home page] or,
-alternatively, in the [quick-start.org] on GitHub.  The particular
-option of `TopW L 90` says to put the bar in the upper left of the
-screen, and make it consume 90% of the width of the screen (we need to
-leave a little bit of space for `trayer-srg`).  If you're up for it—and
-this really requires more shell-scripting than Haskell knowledge—you can
-also try to seamlessly embed trayer into xmobar by using
-[trayer-padding-icon.sh] and following the advice given in that thread.
+position options are documented well in xmobar's [quick-start.org].  The
+particular option of `TopW L 90` says to put the bar in the upper left
+of the screen, and make it consume 90% of the width of the screen (we
+need to leave a little bit of space for `trayer-srg`).  If you're up for
+it—and this really requires more shell-scripting than Haskell
+knowledge—you can also try to seamlessly embed trayer into xmobar by
+using [trayer-padding-icon.sh] and following the advice given in that
+thread.
 
 In the commands list you, well, define commands.  Commands are the
 pieces that generate the content to be displayed in your bar.  These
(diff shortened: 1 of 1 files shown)


liskin:
@jaor look at the diff :-(

liskin:
Maybe it's time to resurrect the xmobar.org website again and use a markdown renderer which generates permalink anchors…

jaor:
On Tue, Jul 12 2022, Tomáš Janoušek wrote:

> @jaor look at the diff :-(

oops, sorry about that.

liskin:
Well, in the absence of confirmation that there will ever be a better render of the docs, I've pushed a Python script to keep those links up to date. Unfortunately I've run out of steam (and time) before automating it in GHA. Not sure if worth it though. :-/

jaor:
On Thu, Jul 14 2022, Tomáš Janoušek wrote:

> Well, in the absence of confirmation that there will ever be a better
> render of the docs, I've pushed a Python script to keep those links up
> to date.

i don't know what happened, but my full answer to your initial comment
seems lost.  i said something along the lines of: 

moving back to markdown (you mention using a markdown parser) sounds
like a lot of work for xmobar, which just recently moved to org.
there's a (cumbersome) way to tell the org parser what #-label to use

```org
* this section
:PROPERTIES:
CUSTOM_ID: "section-label"
:END:

Blah blah

```

should insert a #section-label anchor in the html.  it is also possible
to put a explicit mark there, something like

```org
* this section
<<section-label>>Blah blah

```

we have dozens of sections in xmobar's docs, so doing the above for all
of them is again a lot of work, but perhaps we could just add labels
selectively to the sections referenced in this patch, which are just a
handful.  i have no problem in doing it myself.

would that be better for you/xmonad?

liskin:
Oh, the markdown/org thing is just my ignorance and bad memory. Sorry! :-)

If we can create anchors manually that totally works for us I think.

jaor:
On Fri, Jul 15 2022, Tomáš Janoušek wrote:

> Oh, the markdown/org thing is just my ignorance and bad memory. Sorry! :-)
>
> If we can create anchors manually that totally works for us I think.

okay. i've tried to add some (all?) with this commit:

  https://codeberg.org/xmobar/xmobar/commit/f7391782e8f0e963d488eef9131200c9ca1d6349

it seems to be working in that the fragments of the URLs now are the as
they were in github for those sections (of course, the main repo URL has
to change, but thankfully we're not changing hosting services that often
:)).

is that good? did i miss any?

slotThe:
@jaor these should be all for now, thank you!

jaor:
On Tue, Jul 12 2022, Tomáš Janoušek wrote:

> Maybe it's time to resurrect the xmobar.org website again and use a
> markdown renderer which generates permalink anchors…

the vanilla org renderer in emacs knows how to do that too, but using
explicit additional properties in the sections, which would need to be
(tediously) added.  moving back to markdown for xmobar's docs seems too
much of a hassle, although i guess pandoc would do most of the work for
us.

that said, it's not like we're changing our hosting site every day :)

PR #426, opened 2022-11-02
improve Tall

damhiya:
### Description

I slightly modified `Tall` layout algorithm, so that let xmonad does not draw windows with zero area. It takes effect when the split ratio is exactly 0 or 1.
However I'm not sure whether this change will be welcomed by all users, so please comment about it.

### Checklist


slotThe (commented):
What's the use-case for this that isn't already covered by switching to `Full`?  

Regardless, I don't think this should break anything, but @geekosaur might know better.

damhiya:
It's not for practical purpose. But without this change, it looks bad when gaps and borders are enabled.
screenshot
Please look at the borders on the left

geekosaur (approved):
I don't think this will break anything.

slotThe:
I tried to come up with a property for this, but I'm not very happy with it :/

tests/Properties/Layout/Tall.hs
```diff
@@ -73,7 +73,7 @@ prop_purelayout_tall n r1 r2 rect = do
       st = fromJust . stack . workspace . current $ x
       ts = pureLayout layout rect st
   return $
-        length ts == length (index x)
+        length ts <= length (index x)
```
damhiya: I noticed that this property which does not holds for new `Tall` layout passes the test. I loosened the condition, but this is too weak. It will be good if there is more fine property for `pureLayout` of `Tall`

slotThe: Yes this is definitely too weak.  You should restore `==` for "normal" values of `tallRatio` and look into what would be a good property for `0` and `1`.

The reasons the tests pass is that it's quite unlikely for a `Rational` to come out as `0` or `1`—perhaps this would even warrant a new type with a custom `Arbitrary` instance.  Manually triggering the test; i.e., writing something like

``` haskell
generate $ prop_purelayout_tall 10 (3/100) 0 (Rectangle 0 0 4 8)
```

will fail all (most) of the time.

damhiya:
I fixed the test code for `pureLayout`.

slotThe (approved):
LGTM now!

slotThe:
Thanks!

PR #427, opened 2022-11-03
Fix compatibilities with mtl 2.3.1

a5ob7r:
### Description

No longer mtl version 2.3.1 re-exports Control.Monad, Control.Monad.Fix and Data.Monoid modules, so we need to import them directly instead.

I confirmed this fix by running the command below.

```sh
cabal build --flags=pedantic --constraint='mtl >= 2.3.1'
```

### Checklist


slotThe:
Thanks!

a5ob7r:
Thank you for a quick reply too!

liskin:
I also published a Hackage revision adding < 2.3 to the constraint for mtl, for all of 0.15, 0.17.0 and 0.17.1.
We should do another release before GHC 9.6 comes out.

PR #428, opened 2022-11-05
Fix compatibilities with unix 2.8.0.0

a5ob7r:
### Description

The version of unix removes the 3rd Maybe FileMode argument of openFd, so we need to handle this breaking change using a MIN_VERSION macro. The argument is integrated with the OpenFileFlags argument and the integrated value in defaultFileFlags is Nothing, so there's no difference between the two function calls.

I tested this change by running the command below.

```sh
cabal build --flags=pedantic --constraint='unix >= 2.8'
```

### Checklist

### DIFF
--- a/src/XMonad/Core.hs
+++ b/src/XMonad/Core.hs
@@ -1,3 +1,4 @@
+{-# LANGUAGE CPP #-}
 {-# LANGUAGE DeriveTraversable #-}
 {-# LANGUAGE ExistentialQuantification #-}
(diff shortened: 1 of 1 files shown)


slotThe:
Thank you!

liskin:
I also published a Hackage revision adding `< 2.8` to the constraint for `unix`, for all of 0.15, 0.17.0 and 0.17.1.
We should do another release before GHC 9.6 comes out.

PR #430, opened 2022-11-19
Drop support for GHC 8.4

slotThe:
### Description

Now that Debian has 8.6, it's time to drop 8.4.

##### 7871950 ci: Update to GHC 9.2.5

+ A new stackage LTS is out with GHC 9.2.5, so test for this.
+ Update the haskell-ci jobs accordingly from 9.2.4.

##### 5c7c280 ci: Drop support for GHC 8.4

Debian stable is not on 8.6, which was always our guide as to which GHC
versions we want to support.

In particular, this lets us get rid of all the quickcheck-classes
special treatment.

Related: https://github.com/xmonad/xmonad/commit/400730fe3b8aab8e7f9403215c88b37e44fe7a66

##### cd86480 Use DerivingVia for obvious instances

Certain instance definitions are so automatic, they should be derivable.
Starting with GHC 8.6, they are!

### Checklist

        manually, ...) and concluded: compiles :)

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -66,11 +66,6 @@ jobs:
             compilerVersion: 8.6.5
             setup-method: hvr-ppa
             allow-failure: false
-          - compiler: ghc-8.4.4
-            compilerKind: ghc
-            compilerVersion: 8.4.4
-            setup-method: hvr-ppa
-            allow-failure: false
       fail-fast: false
     steps:
       - name: apt
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -8,9 +8,9 @@
 #
 # For more information, see https://github.com/haskell-CI/haskell-ci
 #
-# version: 0.15.20221009
+# version: 0.15.202211107
 #
-# REGENDATA ("0.15.20221009",["github","cabal.project"])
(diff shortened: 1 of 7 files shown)


LSLeary:
The instances can be simplified a bit further, including xmonad/xmonad-contrib#778. In particular, you don't need to duplicate the inner type for `Ap`, and can just write:

```haskell
newtype Foo a = Foo{ ... }
  deriving newtype (Functor, Applicative, ...)
  deriving (Semigroup, Monoid) via Ap Foo a
```

... Though it seems I was a bit slow. Well, it's not a big deal.

liskin:
Oh, sorry, I'm unfortunately not up to date on all these modern Haskelly things. @slotThe please just push a commit with this change, IMO it is worth it.

slotThe:
> The instances can be simplified a bit further, including [xmonad/xmonad-contrib#778](https://github.com/xmonad/xmonad-contrib/pull/778). In particular, you don't need to duplicate the inner type for `Ap`, and can just write:
>
> ```haskell
> newtype Foo a = Foo{ ... }
>   deriving newtype (Functor, Applicative, ...)
>   deriving (Semigroup, Monoid) via Ap Foo a
> ```
>
> ... Though it seems I was a bit slow. Well, it's not a big deal.

Oh, indeed.  I didn't think of this, but it make perfect sense given how `Ap` is defined and `DerivingVia` works; thanks!

PR #435, opened 2022-12-19
Simplify the design of the manage hook

LSLeary:
### Description

Based on `mvdt`/#432.

##### MVDT: Simplify the design of the manage hook

> `ManageHook` as `Query (Endo WindowSet)` is a holdover from The Old Way.

> `Query` wraps the `X` Monad, which already collects and composes changes
> to the windowset. As such, the `Endo WindowSet` is redundant. Worse, the
> way these changes compose is not at all obvious or intuitive.

> Suppose you compose two ManageHook values:

> ```haskell
> manageHook = (qF $> Endo f)
>           <> (qG $> Endo g)
> ```

> The result of running the `manageHook` on some `w` is not, as one might
> guess,

> ```haskell
> do
>   runQuery qF w
>   windows f
>   runQuery qG w
>   windows g
> ```

> nor is it

> ```haskell
> do
>   runQuery qF w
>   runQuery qG w
>   windows f
>   windows g
> ```

> Unfortunately, it is:

> ```haskell
> do
>   runQuery qF w
>   runQuery qG w
>   windows g
>   windows f
> ```

> This is a wart, and it's one we can now excise.

> Hence the `ManageHook` type synonym is simplified to `Query ()`, and
> `manage` runs the hook as-is. It also performs the initial handling of
> the window beforehand, so it's fully visible within the hook. As such,
> `willFloat` (which repeats the indirect logic in `manage`) is deprecated
> in favour of `isFloat`, which directly references the model.

> Note that this change also allows cleaner syntax in configuration. E.g.

> ```haskell
> manageHook = composeAll
>   [ test1 --> fooHook
>   , test2 --> barQueryAction >> idHook
>   , test3 --> fooHook <> bazHook <> quuxHook
>   ]
> ```

> can become

> ```haskell
> manageHook = do
>   test1 --> fooHook
>   test2 --> barQueryAction
>   test3 --> do
>     fooHook
>     bazHook
>     quuxHook
> ```

> To accomodate the change, `doF` must become a lifted `windows`, but that
> and monoid-polymorphism take care of adjusting the rest of the core
> interface transparently.

> As such, simple configs will continue to compile, but extensions that
> directly interfered with the old `Endo WindowSet` won't.

> Further, due to `doF`s now composing forwards rather than backwards,
> many user configs will break silently. The solution is to reverse the
> order that `doF`, `doFloat`, `doIgnore` and `doShift` appear in,
> extending that list to include anything else built on `doF`, like
> `doShiftTo` or `doFullFloat` from `X.H.ManageHelpers`.

#### Commentary

  - An early incarnation of these changes was tested for some months.

  - Unlike the other PRs, I'm not actually sure this one is advisable. It may be "morally correct", but the degree of config breakage may just be too high for it to be worth it.

### Checklist

        manually, ...) and concluded: they function as intended.

### DIFF
--- a/src/XMonad/Core.hs
+++ b/src/XMonad/Core.hs
@@ -39,6 +39,7 @@ module XMonad.Core (
     ManageHook, Query(..), runQuery, Directories'(..), Directories, getDirectories,
   ) where
 
+import XMonad.Internal.Core (Internal)
 import XMonad.StackSet hiding (modify)
 
 import Prelude
@@ -106,6 +107,7 @@ data XConf = XConf
                                       -- the event currently being processed
     , currentEvent :: !(Maybe Event)  -- ^ event currently being processed
     , directories  :: !Directories    -- ^ directories to use
+    , internal     :: !(Internal WindowSet) -- ^ a hiding place for internals
     }
 
 -- todo, better name
--- a/src/XMonad/Core.hs
+++ b/src/XMonad/Core.hs
@@ -176,7 +176,7 @@ newtype X a = X (RWST XConf Any XState IO a)
 instance Default a => Default (X a) where
     def = return def
 
-type ManageHook = Query (Endo WindowSet)
+type ManageHook = Query ()
 newtype Query a = Query (ReaderT Window X a)
     deriving (Functor, Applicative, Monad, MonadReader Window, MonadIO)
     deriving (Semigroup, Monoid) via Ap Query a
--- a/src/XMonad/Core.hs
+++ b/src/XMonad/Core.hs
@@ -47,10 +47,9 @@ import qualified Control.Exception as E
 import Control.Applicative ((<|>), empty)
 import Control.Monad.Fail
 import Control.Monad.Fix (fix)
-import Control.Monad.State
+import Control.Monad.RWS
 import Control.Monad.Reader
 import Control.Monad (filterM, guard, void, when)
-import Data.Semigroup
 import Data.Traversable (for)
 import Data.Time.Clock (UTCTime)
 import Data.Default.Class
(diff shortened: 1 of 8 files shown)


slotThe (commented):
> Unlike the other PRs, I'm not actually sure this one is advisable. It may be "morally correct", but the degree of config breakage may just be too high for it to be worth it.

This is what I'm thinking as well.  Seems like this would break half of contrib (not to mention user configs); probably not advisable at this point

TheMC47:
> > Unlike the other PRs, I'm not actually sure this one is advisable. It may be "morally correct", but the degree of config breakage may just be too high for it to be worth it.
> 
> This is what I'm thinking as well. Seems like this would break half of contrib (not to mention user configs); probably not advisable at this point

I second this, this seems like a big change where it's hard to predict how things might break. However, I'm gonna try and advocate for merging this at some point:
- For contrib modules: the build will break and we can decide on a case-by-case basis how to fix the problems.
- For user configs:
  - Many configs are just copy-pasted from other places. This can help us identify the patterns that break and provide alternatives and fixes. There are likely just four or five such configs that are just copied _everywhere_.
  - People who write their own configs probably don't update xmonad (we don't really have a  predictable release cycle) or are Haskell-savvy enough to navigate the changes.

I'm not a big fan of guaranteeing backward compatibility at all costs, especially if the cost is disregarding a better alternative. Perfect software doesn't exist, so things will break at some point. However, We need to be careful and mindful on how and when to break things

geekosaur:
We did get a number of complaints about backward compatibility breaking after 0.17.0 came out, though. I'm tempted to say we should schedule a 1.0 release at some point, then save this for the 2.0 series. Possibly the same for the other PRs, since I know I've pointed lots of people to `windows`.

TheMC47:
> we should schedule a 1.0 release at some point, then save this for the 2.0 series

This sounds like a reasonable way to go forward

src/XMonad/Core.hs
```diff
@@ -157,22 +158,25 @@ newtype ScreenDetail = SD { screenRect :: Rectangle }
...
     deriving (Semigroup, Monoid) via Ap X a
 
 instance Default a => Default (X a) where
     def = return def
 
-type ManageHook = Query (Endo WindowSet)
+type ManageHook = Query ()
 newtype Query a = Query (ReaderT Window X a)
     deriving (Functor, Applicative, Monad, MonadReader Window, MonadIO)
     deriving (Semigroup, Monoid) via Ap Query a
```
aavogt, line 182: > Further, due to doFs now composing forwards rather than backwards,
many user configs will break silently. 

You could keep the old order by making it `Dual (Ap Query a)`, but then it becomes even harder to learn.

PR #436, opened 2022-12-19
Purify the unclean

LSLeary:
### Description

Based on `mvdt`/#432.

##### Purify the unclean by generalising type signatures

> Many operations were previously impure due to direct or indirect use of
> the old `windows`, now `render`. Due to MVDT, we can now purify them.

> To do so, we generalise their type signatures away from `X` to
> arbitrary monads satisfying the relevant mtl constraints. This has
> further advantages in terms of code reuse (e.g. in `X.U.PureX`) and
> possibly in testing.

> However, it also causes some breakage—bindings that previously type
> checked without a signature may now need `FlexibleContexts` to do so.

##### Deprecate `runOnWorkspaces` in favour of `traverseWorkspaces`

> `runOnWorkspaces` was a questionable existence in various ways:

>   * It traversed the workspaces in a unexpected, nonstandard order.

>   * It needlessly utilised a non-exhaustive pattern match, abusing
>     first `MonadFail` then irrefutability to evade warning.

>   * It was used nowhere in contrib, once in core—that sole use with a
>     pure function.

>   * Even after generalisation, it still has a bad type, both misleading and
>     lacking in parametricity. Specialising the `StackSet` traversal to the
>     `WindowSet` in `XState` gives the function too much power, and suggests
>     that it's using that power to provide special support for `WindowSet`
>     modifications when in fact it does no such thing.

> To resolve these issues, the workspace traversal logic is written as
> `traverseWorkspaces` in `X.StackSet`, and `mapWorkspaces` written atop
> that via `Identity`. The latter then replaces the sole use of
> `runOnWorkspaces`, condemning it to deprecation.

#### Commentary

  - The first commit has been tested for some months. The second is much newer, but has been running on my system with no issue.

  - The second commit is another kind of hygiene. It arose on account of ugliness I was forced to confront in writing the first, but beyond that, it's not strictly related. It could be split off, but it didn't seem to deserve a PR of its own.

### Checklist

        manually, ...) and concluded: they appear to function as intended.

### DIFF
--- a/src/XMonad/Operations.hs
+++ b/src/XMonad/Operations.hs
@@ -28,7 +28,7 @@ module XMonad.Operations (
     setTopFocus, focus, isFixedSizeOrTransient,
 
     -- * Manage Windows
-    windows, respace, refresh, norefresh, handleRefresh, rescreen,
+    windows, respace, refresh, norefresh, handleRefresh, rendered, rescreen,
     modifyWindowSet, windowBracket, windowBracket_,
     clearEvents, getCleanedScreenInfo,
     withFocused, withUnfocused,
@@ -63,6 +63,7 @@ module XMonad.Operations (
 import XMonad.Core
 import XMonad.Layout (Full(..))
 import qualified XMonad.StackSet as W
+import XMonad.Internal.Operations (rendered, unsafeLogView)
 
 import Data.Maybe
 import Data.Monoid          (Endo(..),Any(..))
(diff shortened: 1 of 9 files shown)


src/XMonad/Operations.hs
```diff
@@ -604,33 +608,33 @@ replace dpy dflt rootw = do
...
 
 -- | Return workspace visible on screen @sc@, or 'Nothing'.
-screenWorkspace :: ScreenId -> X (Maybe WorkspaceId)
+screenWorkspace :: MonadState XState m => ScreenId -> m (Maybe WorkspaceId)
 screenWorkspace sc = withWindowSet $ return . W.lookupWorkspace sc
 
 -- | Apply an 'X' operation to the currently focused window, if there is one.
-withFocused :: (Window -> X ()) -> X ()
+withFocused :: MonadState XState m => (Window -> m ()) -> m ()
 withFocused f = withWindowSet $ \w -> whenJust (W.peek w) f
```
slotThe, line 616: E.g.

``` haskell
withFocused :: MonadState XState m => (Window -> m ()) -> m ()
withFocused f = withWindowSet $ \w -> whenJust (W.peek w) f
{-# SPECIALISE withFocused :: (Window -> X ()) -> X () #-}
```

src/XMonad/StackSet.hs
```diff
@@ -517,6 +520,24 @@ delete' w s = s { current = removeFromScreen        (current s)
...
+  :: Applicative f
+  => (Workspace i l a        -> f (Workspace i' l' a))
+  ->  StackSet  i l a sid sd -> f (StackSet  i' l' a sid sd)
+traverseWorkspaces f s = StackSet
+  <$>          onScreen (current  s)
+  <*> traverse onScreen (visible  s)
+  <*> traverse f        (hidden   s)
+  <*> pure              (floating s)
+  where onScreen scr = f (workspace scr) <&> \w -> scr{ workspace = w }
+
```
slotThe, line 540: Maybe we should reconsider https://github.com/xmonad/xmonad/pull/357 after all...

slotThe (approved):
Maybe this is unwise, but I'm taking an "assume the other PR was already merged" approach to this review (and to all others that follow) :)

All-in-all this looks fine to me.  The only thing I'd change is to add `SPECIALISE` pragmas to the functions that are now generalised (see below), since mtl performance is heavily dependent on whether inlining/specialisation can take place (this may or may not matter, but specialising comes at basically no cost for us).

Issue #373, opened 2022-02-09
More aggressively suggest `ghcup` in INSTALL.md


slotThe:
We are currently suggesting to use ghcup in INSTALL [here](https://github.com/xmonad/xmonad/blob/master/INSTALL.md?plain=1#L138) and [here](https://github.com/xmonad/xmonad/blob/master/INSTALL.md?plain=1#L224).  However, this is done mainly for completeness and I doubt anyone would actually look at ghcup based on those two short sentences.  For distributions where installing statically linked Haskell binaries is semi non-trivial (*cough*), however, using ghcup might be by far the easiest option, instead of installing some statically linked stack binary.  I think we should push for that more (I personally have never used ghcup, so I doubt I'm a good person to write that part of the docs)

harjotgill:
I just spend an hour dealing with compilation errors with stack and cabal methods. The GHCup method is indeed the quickest/easiest way to get started right now.

slotThe:
On Wed, Apr 13 2022 18:03, Harjot Gill wrote:
> I just spend an hour dealing with compilation errors with stack and
> cabal methods. The GHCup method is indeed the quickest/easiest way to
> get started right now.

As always: contributions very welcome!

PR #441, opened 2023-03-13
Suggest GHCup more intensively in INSTALL.md

Equwece:
### Description

Suggest GHCup more intensively in INSTALL.md. Related to #373.

### Checklist

    I'm not sure, but I think it's not necessary for documentation update.

     I'm not sure, but I think it's not necessary for documentation update.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -123,7 +123,9 @@ Unless you already know which one you prefer, use Stack, which is easier.
 
 #### Install Stack
 
-The easiest way to get [stack] is probably via your system's package
+Probably one of the best ways to get [stack] is to use [GHCup], which is the main Haskell installer according to language's official [website][GHCup] and community [survey]. GHCup is [widely available] and is considered less error prone than other installation options.
+
+You can also use your system's package
 manager:
 
 ``` console
@@ -144,9 +146,6 @@ stack via its [documentation][stack]):
 ``` console
 $ curl -sSL https://get.haskellstack.org/ | sh
 ```
-
-Yet another way would be via [ghcup]; this is similar to installers like
-`rustup`, in case you prefer that.
 
 #### Create a New Project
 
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -222,7 +221,9 @@ be that you don't have the required C libraries installed.  See
 
 #### Install cabal-install
 
-The easiest way to get [cabal-install] is probably via your system's package
+Probably one of the best ways to get [cabal-install] is to use [GHCup], which is the main Haskell installer according to language's official [website][GHCup] and community [survey]. GHCup is available for all Linux distributions and is considered less error prone than other installation options.
```
slotThe: We might want to note that GHCup also works on the BSDs, which is relevant for XMonad

slotThe (approved):
Looks fine to fine to me as a whole, thanks!

INSTALL.md
Equwece: The GHCup [website](https://www.haskell.org/ghcup/install/#supported-platforms) says that only FreeBSD officially supported so far. Work on other major BSDs is in [progress](https://github.com/haskell/ghcup-hs/issues/707) as I can see. Should I specify BSD-based operating systems in general or only FreeBSD?

Equwece: Maybe just add a link to the GHCup supported platforms page, like [this](https://github.com/xmonad/xmonad/pull/441/commits/2adeea6e2883d35d2002dc3dabc633c5221ed675)?

liskin: How about "ghcup is widely available" with "widely available" being a link to that?

Equwece: Great, thanks! [Updated](https://github.com/xmonad/xmonad/pull/441/commits/064d5fac5222df449dc26277a0247ab27e6742ca).

slotThe:
Thank you!

PR #446, opened 2023-04-02
X.Operations: Apply size hints in floatLocation

liskin:
### Description

This is hopefully harmless with well-behaved clients and ordinary xmonad configs, and it enables using re-floating to apply size hints to existing windows. The only visible behaviour change I can foresee is that tiled windows (which ignore hints by default, unless X.L.LayoutHints is used) will now snap to size hints whenever they're floated (either via a keybinding or on mouseMoveWindow), whereas previously they'd only do so on mouseResizeWindow.

My use-case for this is the following: when I change fonts in my terminal, it updates its size hints and then sends a ConfigureRequestEvent to change its size to keep the number of rows and columns the same, and it also happens to reset the position to 0, 0. If it's tiled, that request is just ignored and hintsEventHook handles the layout refresh. If it happens to be floating, I want neither the move to 0, 0 nor the window size change to keep rows/colums, so I have a handleEventHook that ignores that ConfigureRequestEvent and just refloats the window, but I need a way to reapply size hints.

I could add a separate function that applies these hints to the floating RationalRect, but that seems like a lossy operation due to the Doubles in there. So I'd probably end up replicating most of the code from floatLocation, and then I might just improve that instead… 🙂

(I'll submit that custom ConfigureRequestEvent-ignoring hook to xmonad-contrib later.)

### Checklist

  - [ ] I've confirmed these changes don't belong in xmonad-contrib instead:  
        see the 3rd paragraph 🙂

  - [X] I've considered how to best test these changes (property, unit, manually, ...) and concluded:  
        Making sure this doesn't break anything involves using it with all sorts of X clients. I've been running this for about a week with no trouble, but I haven't run any Steam games or Libreoffices during this time, so it definitely needs more testing.

  - [x] I updated the `CHANGES.md` file - **TODO** I don't know what to say, yet.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -14,6 +14,10 @@
 
 * `Tall` does not draw windows with zero area.
 
+* `XMonad.Operations.floatLocation` now applies size hints. This means windows
+  will snap to these hints as soon as they're floated (mouse move, keybinding).
+  Previously that only happened on mouse resize.
+
 ### Bug Fixes
 
 * Duplicated floats (e.g. from X.A.CopyToAll) no longer escape to inactive
--- a/src/XMonad/Operations.hs
+++ b/src/XMonad/Operations.hs
@@ -704,6 +704,7 @@ floatLocation w =
 
   where go = withDisplay $ \d -> do
           ws <- gets windowset
+          sh <- io $ getWMNormalHints d w
           wa <- io $ getWindowAttributes d w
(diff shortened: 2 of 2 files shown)


slotThe (approved):
FWIW I've been trying this out for the past few days, and I did not notice any change in behaviour at all (and we did find out that you and I have very different workflows!) 

Then again, I almost aggressively avoid floats, so there's that :)

PR #463, opened 2023-07-22
TUTORIAL: Some Bits

slotThe:
#### TUTORIAL: Add a subsection about the startupHook

This is especially relevant for people who use display managers and thus probably do not want to use .xinitrc directly.

### Checklist

  - [n/a] I've considered how to best test these changes (property, unit,

  - [n/a] I updated the `CHANGES.md` file

### DIFF
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -77,6 +77,11 @@ a live xmonad session in some capacity.  If you have set up your
 `~/.xinitrc` as directed in the xmonad guided tour, you should be good
 to go!  If not, just smack an `exec xmonad` at the bottom of that file.
 
+In particular, it might be a good idea to set a wallpaper beforehand.
+Otherwise, when switching workspaces or closing windows, you might start
+seeing "shadows" of windows that were there before, unable to interact
+with them.
+
 ## Installing Xmobar
 
 What we need to do now—provided we want to use a bar at all—is to
--- a/TUTORIAL.md
+++ b/TUTORIAL.md
@@ -906,10 +906,10 @@ apt-get install nm-applet feh xfce4-power-manager
 
 First, configure xscreensaver how you like it with the
 `xscreensaver-demo` command.  Now, we will set these things up in
-`~/.xinitrc` (we could also do most of this in xmonad's `startupHook`,
-but `~/.xinitrc` is perhaps more standard).  If you want to use xmonad
-with a desktop environment, see [Basic Desktop Environment Integration]
-for how to do this.
+`~/.xinitrc`.  If you want to use XMonad with a desktop environment, see
+[Basic Desktop Environment Integration] for how to do this.  For a
+version using XMonad's built in functionality instead, see the [next
+section][using-the-startupHook].
 
 Your `~/.xinitrc` may wind up looking like this:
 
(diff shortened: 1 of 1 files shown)


TUTORIAL.md
```diff
@@ -952,6 +957,38 @@ Mission accomplished!
 Of course substitute the wallpaper for one of your own.  If you like the
 one used above, you can find it [here](https://i.imgur.com/9MQHuZx.png).
 
+### Using the `startupHook`
```
geekosaur, line 960: I wonder if it's worth mentioning [fbautostart](https://github.com/paultag/fbautostart) here. It appears to have originated within the Ubuntu project, so it may not be packaged for other distributions.

geekosaur (approved):

TUTORIAL.md
slotThe: Mh, the README doesn't say much; what exactly does it do and how does one use it?

geekosaur: Yes, the documentation is sadly sparse. It enables use of XDG autostart programs (`*.desktop` files in `/etc/xdg/autostart` and `~/.config/autostart`) without running a full desktop manager. For xmonad you would `spawnOnce` it in `startupHook`.

geekosaur: I suppose we should also in that case mention `lxsession-edit` as a way to create and manage them without a full desktop manager.

geekosaur: Actually it looks like just "manage"; unlike the MATE version, you can't create them. Sadly the MATE one isn't standalone so I can't suggest using that with xmonad.

geekosaur: Also I should mention that the main reason you would do this is that there are applications that automatically install XDG autostart entries, so you would either need to translate those into `startupHook` entries or arrange for XDG autostarts to run for them to work properly. In my case (xmonad under MATE) I don't have to worry about this, since MATE does it, but people who don't use xmonad with a desktop manager may care.

geekosaur: Looks like there's also `xdg-autostart` from the `obsession` package, which goes with Openbox by default. But, unlike `fbautostart`, already supports use with other environments (XDG allows you to specify that startup programs can be associated with specific environments instead of always started.) This one might be more generally available, whereas the other seems to have been written specifically for Ubuntu.

slotThe: What are these programs that automatically install XDG autostart entries? I've never seen this before. Also, I can't find any `obsession` package in the usual places (Nix, Arch, … repositories); perhaps that is again something Debian/Ubuntu specific?

geekosaur: Google's telling me Debian and Slackware. It looks like `fbautostart` is in AUR, so it may be the more common one. The Arch Wiki also suggests a `dex` package (https://github.com/jceb/dex), which also contains a full editor for them.

The programs that contain autostart entries are generally intended for full desktop environments (mainly Gnome) so maybe this isn't so relevant.

slotThe: Mh, yeah it's perhaps a bit outside of the scope of this PR. It'd be nice if *someone* wrote something relating all of these different programs, but I don't feel like I have the expertise to do so

PR #477, opened 2023-10-24
ci: Regenerate haskell-ci

liskin:


slotThe:
I reckon we should disable erroring on `Wx-partial` for tests

PR #482, opened 2023-12-09
ci: Regenerate haskell-ci

liskin:


liskin:
Hm, changes like these are just useless noise, I should find a way to avoid PRs that only change comments.

liskin:
> Hm, changes like these are just useless noise, I should find a way to avoid PRs that only change comments.

Done: https://github.com/liskin/xmonad-maintenance/commit/c4cfded1a2af12bf1483a33e2ae78bd75d0756f5

slotThe:
Nice, thank you!

PR #511, opened 2024-07-29
Revert #404

geekosaur:
### Description

The join point bug in ghc 9.0 and early 9.2 has hopefully run its course: nobody should be using those versions any more.

### Checklist

        manually, ...) and concluded: not needed (or possible unless someone's still using 9.2.1)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -8,6 +8,16 @@
 
 ### Bug Fixes
 
+### Other
+
+PR #404 (see last change in 0.17.1) has been reverted, because the affected
+compilers are (hopefully) no longer being used.
+
+All 9.0 releases of GHC, plus 9.2.1 and 9.2.2 have the join point bug.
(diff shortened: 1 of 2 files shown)


geekosaur:
I suggest squashing this. (ETA: squashed manually.)

slotThe (approved):
Looks fine to me :+1: I have not checked to see if the crash still happens, but it's true that https://gitlab.haskell.org/ghc/ghc/-/merge_requests/8597 was merged a while ago

geekosaur:
9.2.3 and all later releases have the fix. 9.0.x doesn't, but the official word from GHC HQ is that 9.0 is too buggy to use (granting that the join point bug is one of the reasons, but there are also significant GC bugs IIRC).

geekosaur:
Added a little more detail to the changelog about affected and recommended ghc versions.

PR #520, opened 2024-12-15
Initialize XMONAD_* env variables separatly

Profpatsch:
The user shouldn’t have to input all three variables if they just want to set where e.g. the config is.

This tries to achive that, but of course we need to keep the heuristics for the `xmonad` config dir, so set the data dir here if any of the config/cache files exist and it was not overwritten.

### Checklist

### DIFF
--- a/src/XMonad/Core.hs
+++ b/src/XMonad/Core.hs
@@ -523,35 +523,61 @@ type Directories = Directories' FilePath
 -- always assumed to be in @cfgDir@.
 --
 getDirectories :: IO Directories
-getDirectories = xmEnvDirs <|> xmDirs <|> xdgDirs
+getDirectories = do
+    dataDir <- mma xmEnvDataDir xmDataDir xdgDataDir
+    cfgDir <- mma xmEnvCfgDir xmCfgDir xdgCfgDir
+    cacheDir <- mma xmEnvCacheDir xmCacheDir xdgCacheDir
+    pure $ Directories { dataDir, cfgDir, cacheDir}
+
   where
+    mma fa fb fc = fa >>= \case
+        Just a -> pure a
+        Nothing -> fb >>= \case
+          Just b -> pure b
+          Nothing -> fc
+
     -- | Check for xmonad's environment variables first
-    xmEnvDirs :: IO Directories
-    xmEnvDirs = do
-        let xmEnvs = Directories{ dataDir  = "XMONAD_DATA_DIR"
-                                , cfgDir   = "XMONAD_CONFIG_DIR"
-                                , cacheDir = "XMONAD_CACHE_DIR"
-                                }
-        maybe empty pure . sequenceA =<< traverse getEnv xmEnvs
+    xmEnvDataDir, xmEnvCfgDir, xmEnvCacheDir :: IO (Maybe FilePath)
+    xmEnvDataDir  = getEnv "XMONAD_DATA_DIR"
+    xmEnvCfgDir   = getEnv "XMONAD_CONFIG_DIR"
+    xmEnvCacheDir = getEnv "XMONAD_CACHE_DIR"
+
+    -- this is a tiny bit inefficient, but it's not worth the extra complexity
 
     -- | Check whether the config file or a build script is in the
-    -- @~\/.xmonad@ directory
-    xmDirs :: IO Directories
-    xmDirs = do
+    xmDataDir, xmCfgDir, xmCacheDir :: IO (Maybe FilePath)
+    xmCfgDir = do
         xmDir <- getAppUserDataDirectory "xmonad"
         conf  <- doesFileExist $ xmDir </> "xmonad.hs"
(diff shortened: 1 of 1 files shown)


src/XMonad/Core.hs
```diff
@@ -523,35 +523,61 @@ type Directories = Directories' FilePath
...
 
     -- | Check whether the config file or a build script is in the
-    -- @~\/.xmonad@ directory
-    xmDirs :: IO Directories
-    xmDirs = do
+    xmDataDir, xmCfgDir, xmCacheDir :: IO (Maybe FilePath)
+    xmCfgDir = do
         xmDir <- getAppUserDataDirectory "xmonad"
         conf  <- doesFileExist $ xmDir </> "xmonad.hs"
-        build <- doesFileExist $ xmDir </> "build"
```
geekosaur, line 543: We still want to check for a `build` script as an alternative to `xmonad.hs`, and it belongs in the config directory.

geekosaur (approved):

geekosaur (changes requested):
I think the build script change is a blocker for this.

src/XMonad/Core.hs
slotThe: Yes; I think the logic should be something like the following

  1. If any of the `XMONAD_*` env vars are set, use those for the relevant directories, and default to the XDG directories for those that are unset.

  2. Check for `xmonad.hs`/`build` script in the usual places, and set everything to one directory (this is especially important for backwards compatibility)

  3. If all else fails, default to the XDG spec.

liskin: The code would probably become simpler if we retained the xdg/legacy detection intact, and only then applied env var overrides individually. I'm not 100% sure everyone agrees with the semantics that result, though…

…and this not being sure stems mainly from me being unable to understand @slotThe's comment 🙂
You mention defaulting to XDG twice, and I'm confused what that means.

geekosaur: I understood it to mean "if there are no files present in either location, assume XDG". Which should keep things working for folks using `~/.xmonad` while doing what new people these days expect.

geekosaur: I should probably also mention that the reason `build` is an alternative is that someone in the past was using it to point to an `xmonad.hs` in their dotfiles repo; they disapproved of the idea that source files could be under `~/.config`. (My own opinion is that, while I see their point, it is only a small deviation from using a config file in an _ad hoc_ configuration language and you also see this with programs which use e.g. Tcl or Lua as their configuration language.)

Profpatsch: So I check if xmonad/xmonad.hs exists, if so I set to that, if not I fall back to envvars & xdg dirs? I think that’s the easiest to understand and backward compatible.

So 
 - if `~/.xmonad` exists use that
 - if not, use the xdg spec (and allow the envvars) to override
 - in addition, if all three env vars are set always supersede (for backwards compat with current behaviour)

geekosaur: Sounds sensible to me, but we've thought that before and gotten issues filed afterward. Occupational hazard. 😀

geekosaur: Hm, actually, we decided that simply having `~/.xmonad` exist wasn't enough because (IIRC) there are apparently xmonad distro packages that autocreate it. We ended up checking specifically for `xmonad.hs` or `build` as a result, not merely for the directory.

Profpatsch: Is there a way to log warnings? I scrolled the code a bit but didn’t find logging functions

slotThe: > …and this not being sure stems mainly from me being unable to understand @slotThe's comment 🙂 You mention defaulting to XDG twice, and I'm confused what that means.

My point was mostly what to do when, for example, only `XMONAD_CONFIG_DIR` is set. One could then cram everything else (data and cache dirs) into `~/.xmonad`, but I think the XDG dirs are a more natural choice in that case

> So I check if xmonad/xmonad.hs exists, if so I set to that, if not I fall back to envvars & xdg dirs? I think that’s the easiest to understand and backward compatible.

`~/.xmonad/xmonad.hs` or `~/xmonad/build`; but yes, something like that sounds fine I think

geekosaur: xmonad was written to do things as minimally as possible; logging is `hPutStrLn stderr $ whatever` and goes to wherever the session log is.

Profpatsch:
fwiw, the patch kinda solves my problem but the more general issue seems to be quite complicated and I don’t want to step on people’s weird configuration setups. Plus I also don’t want to lose an undue amount of time on this.

So if somebody else wants to adjust the patch to make it work somehow, feel free, but I think I’m gonna abandon this PR and patch downstream for now.

Profpatsch:
Make sure to adjust the manpage as well, because that has a section of how config is handled :)

PR #523, opened 2024-12-21
Add recompilation support via `cabal`

sol:
### Description

Use `cabal` for `--recompile` if there is a `.cabal` file in the config directory and none of `build`, `stack.yaml`, `flake.nix`, nor `default.nix` exist.

- This allows you to e.g. put your source files into `src` instead of `lib`, use `default-extensions`, etc.  As long as your config works with `cabal build` you are good to go.
- `cabal.project` is fully supported, but not required.
- You can split up your config into sub packages or reference GitHub dependencies, etc.

### Checklist

        manually, ...) and concluded: I tested this manually.  Some of the things could benefit from unit tests, but I think we don't have infrastructure for this in place.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -4,6 +4,10 @@
 
 ### Breaking Changes
 
+ * Use `cabal` for `--recompile` if there is a `.cabal` file in the config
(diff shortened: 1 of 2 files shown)


geekosaur (approved):

slotThe:
Thanks!

PR #535, opened 2025-07-12
ci: Regenerate haskell-ci

liskin:

### DIFF
--- a/.github/workflows/haskell-ci.yml
+++ b/.github/workflows/haskell-ci.yml
@@ -203,10 +203,6 @@ jobs:
           echo "packages: ${PKGDIR_xmonad}" >> cabal.project
           echo "package xmonad" >> cabal.project
           echo "    ghc-options: -Werror=missing-methods -Werror=missing-fields" >> cabal.project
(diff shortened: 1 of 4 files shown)


slotThe:
I guess we want to ignore the tests when it comes to some of these warnings. Or remove the new manual insertion of a new package (is this something some haskell-ci now does on its own?)

liskin:
> Or remove the new manual insertion of a new package (is this something some haskell-ci now does on its own?)

Yeah those are new features in haskell-ci:
* https://github.com/haskell-CI/haskell-ci/pull/789
* https://github.com/haskell-CI/haskell-ci/pull/790

We could disable them in [cabal.haskell-ci](https://github.com/xmonad/xmonad/blob/master/cabal.haskell-ci), but it's actually not a bad idea to keep them enabled outside of tests as they pass and seem sensible.

slotThe:
`cabal.project` seems to be horribly global, so I just enabled these checks locally and told `haskell-ci` to stay off it. Perhaps it's better anyways that people can test their changes with the same flags that the CI runs with

geekosaur:
"Horribly"? The whole point of it is to be global; that's what the `.project` means.

slotThe:
Which doesn't end up being what I wanted to do with it here—user error :)

PR #540, opened 2025-10-21
Update cabal install instructions

ffaf1:
### Description

Since `cabal` 3.10 executables are put in `~/.local/bin`.

GHCup recommends 3.14 now so, there is little reason to keep the old location.

### Checklist

        manually, ...) and concluded: no tests.

  - [x] ~~I updated the `CHANGES.md` file.~~ I did not, because the change is trivial and does not modify `xmonad` behaviour. If it is necessary, I will do it.


geekosaur (approved):

slotThe:
Thanks!

PR #542, opened 2025-10-27
Pure layout support

mathstuf:
### Description

Split out a `PureLayout` typeclass. This will allow such layouts to work on abstract windows (e.g., to tile windows on behalf of `river`).

---

Currently looking at trying to improve the `LayoutClass` implementation for `PureLayout` types. Haven't been able to get `deriving via` to work, mostly due to kind mismatches.

### Checklist

### DIFF
--- a/src/XMonad/Core.hs
+++ b/src/XMonad/Core.hs
@@ -169,7 +169,7 @@ newtype ScreenDetail = SD { screenRect :: Rectangle }
 -- instantiated on 'XConf' and 'XState' automatically.
 --
 newtype X a = X (ReaderT XConf (StateT XState IO) a)
-    deriving (Functor, Applicative, Monad, MonadFail, MonadIO, MonadState XState, MonadReader XConf)
+    deriving (Functor, Applicative, Monad, MonadFail, MonadIO, MonadState XState, MonadReader XConf) via ReaderT XConf (StateT XState IO)
     deriving (Semigroup, Monoid) via Ap X a
 
 instance Default a => Default (X a) where
--- a/src/XMonad/Core.hs
+++ b/src/XMonad/Core.hs
@@ -29,6 +29,7 @@ module XMonad.Core (
     X, WindowSet, WindowSpace, WorkspaceId,
     ScreenId(..), ScreenDetail(..), XState(..),
     XConf(..), XConfig(..), LayoutClass(..),
+    Pure(..), PureLayout(..),
     Layout(..), readsLayout, Typeable, Message,
     SomeMessage(..), fromMessage, LayoutMessages(..),
     StateExtension(..), ExtensionClass(..), ConfExtension(..),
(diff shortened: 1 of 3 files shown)


geekosaur:
I don't have a problem with this, but I think pretty much every interesting layout requires access to xmonad internals. I'm wondering if a better approach would be to find a way to expose river's internals as a modified "X monad", presumably without the `StackSet`. This would still leave out some layouts, but accessing anything other than the current `Stack` (which is passed as a parameter) is unusual.

There's also the problem that messages might be rather difficult to support if you're using layouts with a different window manager.

src/XMonad/Core.hs
```diff
@@ -345,6 +369,11 @@ class (Show (layout a), Typeable layout) => LayoutClass layout a where
     description :: layout a -> String
     description      = show
 
+instance (PureLayout layout a) => LayoutClass (Pure layout) a where
```
geekosaur, line 372: Strictly, the parentheses are unnecessary.

```suggestion
instance PureLayout layout a => LayoutClass (Pure layout) a where
```

geekosaur (approved):

src/XMonad/Core.hs
```diff
@@ -256,6 +257,29 @@ data Layout a = forall l. (LayoutClass l a, Read (l a)) => Layout (l a)
 readsLayout :: Layout a -> String -> [(Layout a, String)]
 readsLayout (Layout l) s = [(Layout (asTypeOf x l), rs) | (x, rs) <- reads s]
 
+newtype Pure l a = MkPure (l a)
+    deriving (Show, Typeable)
```
LSLeary, line 261: * `Typeable` is redundant (GHC will derive it anyway).
 * Derive `Read` (for `Layout`).
 * Should probably derive `PureLayout` too.
 * Document purpose/usage.

src/XMonad/Core.hs
```diff
@@ -345,6 +369,11 @@ class (Show (layout a), Typeable layout) => LayoutClass layout a where
     description :: layout a -> String
     description      = show
 
+instance (PureLayout layout a) => LayoutClass (Pure layout) a where
+    pureLayout (MkPure l) = pureLayout' l
+    pureMessage (MkPure l) = fmap MkPure . pureMessage' l
```
LSLeary, line 374: Better to use `coerce` than `fmap MkPure` (may require type ascriptions or applications).

src/XMonad/Layout.hs
```diff
@@ -124,6 +133,11 @@ splitVerticallyBy f = (mirrorRect *** mirrorRect) . splitHorizontallyBy f . mirr
 -- | Mirror a layout, compute its 90 degree rotated form.
 newtype Mirror l a = Mirror (l a) deriving (Show, Read)
 
+instance PureLayout l a => PureLayout (Mirror l) a where
+    pureLayout' (Mirror l) r = pureLayout' l (mirrorRect r)
+    pureMessage' (Mirror l) = fmap Mirror . pureMessage' l
```
LSLeary, line 135: Again, `coerce`.

src/XMonad/Layout.hs
```diff
@@ -124,6 +133,11 @@ splitVerticallyBy f = (mirrorRect *** mirrorRect) . splitHorizontallyBy f . mirr
 -- | Mirror a layout, compute its 90 degree rotated form.
 newtype Mirror l a = Mirror (l a) deriving (Show, Read)
 
+instance PureLayout l a => PureLayout (Mirror l) a where
+    pureLayout' (Mirror l) r = pureLayout' l (mirrorRect r)
```
LSLeary, line 134: This looks wrong. Don't you also need to `mirrorRect` the windows `pureLayout'` gives back?

src/XMonad/Core.hs
```diff
@@ -169,7 +169,7 @@ newtype ScreenDetail = SD { screenRect :: Rectangle }
 -- instantiated on 'XConf' and 'XState' automatically.
 --
 newtype X a = X (ReaderT XConf (StateT XState IO) a)
```
LSLeary, line 171: The point of using `deriving via` here is to allow you to inline the contents of these newtypes (and hence obtain the `type role X representational`); the field should be `XConf -> XState -> IO (a, XState)`.

LSLeary (changes requested):
Needs some changes, otherwise looks fine.

src/XMonad/Core.hs
geekosaur: `Typeable` may have come from existing xmonad code: we still have it in the codebase, for support of older compilers that still require you to in effect declare that you're going to use it. That only changed relatively recently, I think.

LSLeary: https://downloads.haskell.org/ghc/latest/docs/users_guide/using-warnings.html#ghc-flag-Wderiving-typeable

It's been auto-derived since *7.10*.

geekosaur: Yes, but for a number of versions after that you still needed to tell it _you_ wanted to use it, in various ways. I'm not sure why, given it was always present anyway.

mathstuf:
>: I don't have a problem with this, but I think pretty much every interesting layout requires access to xmonad internals.

I see a number of `PureLayout` candidates in `xmonad-contrib`. The `LayoutModifier` stuff is likely also doable. Most of the ineligible ones do things like draggable dividers and the like. There are a few others, but it looks like most of them should support it.

> There's also the problem that messages might be rather difficult to support if you're using layouts with a different window manager.

Good point. I didn't want to embark on a journey for this before knowing the foundations are even plausible. I'll look into messages as well if I can at least get the rectangle juggling understood.

PR #555, opened 2026-01-11
add Hécate's Get Tested

geekosaur:
### Description

Start replacing Haskell-CI, which has been more of a problem than a solution of late, with @Kleidukos's Get Tested!. This is a functional first cut, but will probably want more work in the future since Get Tested! doesn't do some things Haskell-CI does.

### Checklist

        manually, ...) and concluded: everything but Hackage releases tested, obviously;
        someone should test release candidates next time we make a release (see note below)

  - [ ] I updated the `CHANGES.md` file (not user visible, CI only)

### Comments

  - I didn't clean up the generated build steps I copied from Haskell-CI; there's some optimization that could be done.
  - The candidate and release actions are disabled (`if: false &&`) so they don't get triggered inadvertently. Remember to change these if we decide to test or switch to this.

### DIFF
--- a/.github/workflows/get-tested.yml
+++ b/.github/workflows/get-tested.yml
@@ -1,0 +1,345 @@
+name: Get Tested!
+
+on:
+  push:
+  pull_request:
+  merge_group:
+  release:
+    types:
+      - published
+  workflow_dispatch:
+    inputs:
+      version:
+        description: candidate version (must match version in cabal file)
+
+env:
+  GHC_FOR_UPLOAD: 9.8
+  CABAL_VERSION: 3.16.1.0
+  GHCUP_VERSION: 0.1.50.1
+
+jobs:
+  generate-matrix:
+    name: "Generate matrix from cabal file"
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Extract the tested GHC versions
+        id: set-matrix
+        uses: kleidukos/get-tested@v0.1.9.1
+        with:
+          cabal-file: xmonad.cabal
+          ubuntu-version: "24.04"
+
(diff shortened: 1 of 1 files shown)


geekosaur:
Also, this will need to be repeated for the other repos. Sadly, it's not either drop-in or a mechanical change.

slotThe:
IRC conversation, for completeness:

    <Solid> I wonder if this is a necessary change, actually; it seems to me
            that haskell-ci works just fine, we just need to make lliskins
            automatic bumps aware of dependabot
    <liskin> Agreed with Solid, the PR doesn't look like something we want
             to adopt and maintain, unless haskell-ci is truly dead (which
             it is not, last time I checked). And even then, someone else
             would take over most likely, it's not like we have the extra
             resources to maintain it
    <liskin> We could just open PRs to haskell-ci btw...
    <liskin> I wanted to make the auto regen thing be more aware but the
             only point of that is so we don't need to open the PRs within a
             few weeks. If we can open them, that's almost certainly
             better...
    <geekosaur> the big thing I see in get-tested is that it doesn't need a
                cron job, it generates the matrix on the fly
    <geekosaur> and how much of the boilerplate in haskell-ci comes from
                that big patch 🙂
    <liskin> Very little of the boilerplate comes from the big patch.
    <liskin> If anything, we could perhaps make a patch (or something) that
             uses haskell-ci for the boilerplate and then replaces the
             static matrix with get-tested to reduce the yaml churn in git
    <liskin> That sounds like a fun exercise...
    <liskin> And then we still need my automation to keep tested-with in the
             cabal files up to date, btw
    <geekosaur> ah

liskin:
I believe we've addressed most of the annoyances with haskell-ci for now and I'm more or less back to being able to react to urgent stuff so… shall we close this for now?

